PR #1, opened 2015-10-06
Use postgres instead of sqlite

tadast:
We're planning to use postgres for storing things in this application, change sqlite gem to pg, update database.yml (borrowed from [here](https://github.com/alphagov/policy-publisher/blob/c50215bc48a8e5894bd905e58b99a12abf2a5922/config/database.yml)) and add an empty db/schema.rb


AndrewVos:
lgtm

PR #4, opened 2015-10-08
Persistence (don't merge me)

AndrewVos:


tadast:
This is very much a spike with no tests and the code is not nice either, maybe we can keep it in the branch until we're sure this is the right approach and avoid merging to master?

AndrewVos:
Yeah, I suppose.

KushalP:
I don't mind having this PR open and worked on as a way to keep everyone updated on what's happening. However, we should make it clear that it shouldn't be merged by updating the title.

PR #7, opened 2015-10-12
Configure gds-sso

tadast:
Require authentication via signon in order to access anything in this application. I've not set up permission rules for now as we don't know the logic for that yet.

@KushalP can you have a look at this, paying some extra attention to the ops parts: is there any configuration missing, do we need to modify deploy scripts or create a trello task for setting up tokens in signotron2 in all environments?


KushalP:
There's already a story in progress to create the relevant tokens in all environments: https://trello.com/c/q5SC0Eqk

KushalP:
This all looks good to me. One thing I'd add is a comment to the top of `config/initializers/gds-sso.rb` saying that it will be overwritten on deploy.

tadast:
Would it be ok to add the comment you mentioned and push directly to master given that's the only change?

KushalP:
Don't worry about that last comment. We're moving to 12-Factor apps so these can be injected as ENV vars.

PR #8, opened 2015-10-12
Improvements to documentation

tadast:
A very basic README and comments on initialisers that get overwritten on deploy.


KushalP:
Yup, @jamiecobbett is right here. I think this PR is great, but we should remove 37cce18 from it. Can you do that please @tadast?

tadast:
Rebased & force-pushed :)

PR #11, opened 2015-10-14
Allow editing of guide pages

AndrewVos:
MVP for publishing service manual guides. It doesn't follow the publishing process, but works well as a proof of concept of getting a piece of data from the publisher application to the frontend.


app/models/guide.rb
```diff
@@ -0,0 +1,36 @@
...
+  PUBLISHERS = {
+    "Design Community" => "http://sm-11.herokuapp.com/designing-services/design-community/",
+    "Agile Community" => "http://sm-11.herokuapp.com/agile-delivery/agile-community"
+  }
+
+  before_validation on: :create do |object|
+    object.content_id = SecureRandom.uuid
+  end
+
+  after_save :publish
```
tadast: Should we do something like [this](https://github.com/alphagov/policy-publisher/blob/c50215bc48a8e5894bd905e58b99a12abf2a5922/app/controllers/policies_controller.rb#L29) instead? We haven't merged this yet but already have had issues with callbacks in tests https://github.com/alphagov/service-manual-publisher/commit/7211d15e23f9bd51c792aa6d87778f50c7cb51c0.

KushalP: We should aim to separate out the process of saving a piece of content from publishing a piece of content.

They're completely different steps in the workflow.

Gemfile
```diff
@@ -11,6 +11,8 @@ gem 'plek', '~> 1.10'
 gem 'airbrake', '~> 4.2.1'
 gem 'govuk_admin_template', '~> 3.0.0'
 gem 'gds-sso', '~> 11.0.0'
+gem 'gds-api-adapters'
```
KushalP: Whenever we add a dependency we should provide an explicit version: https://github.com/alphagov/styleguides/blob/master/ruby.md#gemfile

AndrewVos:
lgtm

PR #15, opened 2015-10-20
Document data seeding

AndrewVos:
Explain the use of `rake db:seed`.


tadast:
This could be formulated using stronger and more accurate language i.e. avoiding 'some' and 'should'. Otherwise LGTM :+1:

KushalP:
Agreed with what @tadast says here. It might be easier to understand if it introduced each of the steps:

Running `bundle exec rake db:seed` will load sample data into the database. The sample data is taken from the existing (to be deprecated) [government-service-design-manual](https://github.com/alphagov/government-service-design-manual/) repository. If you don't have a local clone of the [government-service-design-manual](https://github.com/alphagov/government-service-design-manual/), running `db:seed` will clone it for you.

PR #17, opened 2015-10-20
Flow basics

tadast:
Add the very basic part of the content editing flow: if the latest editions of a guide is published, save a new edition instead of updating the published one. This essentially means that currently there can be only one work in progress edition draft, once it's published it should never change.

I've also added a list of guides with an action that indicates whether a new edition will be created or an existing one is going to be edited.

screenshot 2015-10-20 12 07 54

I'm not entirely happy with the implementation, I had to create Guide#update_attributes_from_params to ensure we do not update the existing latest_edition when it's in state = 'published'. Simply assigning a new instance of edition to guide.latest_edition doesn't work, because it's a has_one relationship and Rails deletes the old latest_edition record when a new instance is assigned.


AndrewVos:
Maybe add `bullet` in development?

tadast:
I like [bullet](https://github.com/flyerhzm/bullet) too. Let's add it in a separate PR

KushalP:
:+1: for adding bullet to development.

tadast:
Any feedback on the code changes?

app/models/guide.rb
```diff
@@ -13,4 +13,20 @@ class Guide < ActiveRecord::Base
   before_validation on: :create do |object|
     object.content_id = SecureRandom.uuid
   end
+
+  def work_in_progress_edition?
+    latest_edition.try(:state) == 'draft'
+  end
+
+  def update_attributes_from_params(params, state:)
```
KushalP: Is this missing a default for the `state` keyword?

AndrewVos:
lgtm :+1:

app/models/guide.rb
tadast: This makes it a required named parameter: if the client omits `state:`, an `ArgumentError: missing keyword: state` error will be raised, which in my opinion is more descriptive than `ArgumentError: wrong number of arguments`

I could have made them both named, but hesitated thinking this would appear confusing as the first parameter is a hash. What's your thinking?

KushalP: That makes a little more sense.

My thinking was that the underlying value of `state` can only be so many things. We don't want it to be an integer, `nil` or anything else. It should one of `draft` or `publishing`. Maybe we should make that clearer here if that's what you were trying to do?

tadast: No, that was not my reasoning. It's a good point though, but we should use Active Model validations on an edition model to solve this. That's a separate task imho

KushalP: :+1: to both in that case.

PR #19, opened 2015-10-20
Add bullet gem

AndrewVos:
This gem shows any N+1 queries that might happen. It is only installed
in the `development` environment.


tadast:
A duplicate of https://github.com/alphagov/service-manual-publisher/pull/18.

PR #21, opened 2015-10-20
Prepend /service-manual/ to all guide slugs (for discussion, don't merge)

AndrewVos:
https://trello.com/c/viKuy1Z4/23-make-sure-all-published-documents-live-under-the-same-base-path

I'm very much in favour of replacing this entire pull request with:
```ruby
validates_format_of :slug, :with => /^\/service-manual\//
```

At the moment it looks like this:

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -16,6 +16,7 @@ class GuidesController < ApplicationController
 
   def edit
     @guide = Guide.find(params[:id])
+    @guide.slug = @guide.slug.gsub(/^\/service-manual\//, '')
   end
 
   def update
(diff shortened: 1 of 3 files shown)


app/controllers/guides_controller.rb
```diff
@@ -33,6 +34,7 @@ def update
 private
 
   def guide_params
+    params[:guide][:slug] = File.join("/service-manual/", params[:guide][:slug])
     params.require(:guide).permit(
       :slug,
       latest_edition_attributes: [
```
tadast: It's not great what we have in this file I think. Alternatives:

## Validating input

As suggested by Andrew in an offline discussion, we could simply add a validation for the slug format and leave it to the content designer to edit it. Not the greatest UI but a very simplistic solution.

## Not storing the prefix in the local DB

Prepend `/service-manual/` before calling publishing-api. This is not great as a developer joining the project would have to be aware of this quirk, otherwise it could be very confusing.

## Encapsulating this login 

Encapsulate this logic within a model or some sort of presenter object that prepends `/service-manual` on `Edition#slug=` and removes it when displaying in the form. Could end up being a lot of extra code.

@KushalP any other ideas?

PR #22, opened 2015-10-21
Prepend /service-manual/ to all guide slugs

AndrewVos:
In order to avoid conflicts with paths used by other applications, ensure that
each published document slug is prefixed with /service-manual/.

This is an alternative to #21

### DIFF
--- a/spec/models/guide_spec.rb
+++ b/spec/models/guide_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe Guide do
   describe "on create callbacks" do
     it "generates and sets content_id on create" do
       edition = Generators.valid_edition(title: "something", state: "published")
-      guide = Guide.create!(slug: "/slug", content_id: nil, latest_edition: edition)
+      guide = Guide.create!(slug: "/service-manual/slug", content_id: nil, latest_edition: edition)
       expect(guide.content_id).to be_present
     end
   end
@@ -12,7 +12,7 @@ RSpec.describe Guide do
   describe "#update_attributes_from_params" do
     it "updates latest_edition if it's currently draft" do
       edition = Generators.valid_edition(title: "something", state: "draft")
-      guide = Guide.create!(slug: "/slug", latest_edition: edition)
+      guide = Guide.create!(slug: "/service-manual/slug", latest_edition: edition)
 
       guide.update_attributes_from_params({ latest_edition_attributes: edition.attributes.merge(title: "New title") }, state: 'draft')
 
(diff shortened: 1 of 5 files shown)


spec/models/guide_spec.rb
```diff
@@ -4,8 +4,24 @@
...
     it "generates and sets content_id on create" do
       edition = Generators.valid_edition(title: "something", state: "published")
-      guide = Guide.create!(slug: "/slug", content_id: nil, latest_edition: edition)
+      guide = Guide.create!(slug: "/service-manual/slug", content_id: nil, latest_edition: edition)
       expect(guide.content_id).to be_present
     end
   end
+
+  describe "validations" do
+    it "allows 'published' state" do
```
tadast: Description doesn't look right here :)

spec/models/guide_spec.rb
```diff
@@ -4,8 +4,24 @@
...
+      edition = Generators.valid_edition(title: "something", state: "published")
+      edition = Guide.new(slug: "/service-manual/something", latest_edition: edition)
+      edition.valid?
+      expect(edition.errors.full_messages_for(:slug).size).to eq 0
+    end
+
+    it "doesn't allow slugs without /service-manual/ prefix" do
+      edition = Generators.valid_edition(title: "something", state: "published")
+      edition = Guide.new(slug: "/something", latest_edition: edition)
```
tadast: `s/edition/guide/`

spec/models/guide_spec.rb
```diff
@@ -4,8 +4,24 @@
...
+  describe "validations" do
+    it "allows 'published' state" do
+      edition = Generators.valid_edition(title: "something", state: "published")
+      edition = Guide.new(slug: "/service-manual/something", latest_edition: edition)
+      edition.valid?
+      expect(edition.errors.full_messages_for(:slug).size).to eq 0
+    end
+
+    it "doesn't allow slugs without /service-manual/ prefix" do
+      edition = Generators.valid_edition(title: "something", state: "published")
```
tadast: Do we need this extra line? Can't we do 
```
Guide.new(slug: "/something", latest_edition: Generators.valid_edition)
```
?

app/views/guides/_form.html.erb
```diff
@@ -4,7 +4,7 @@
       <legend>Meta</legend>
       <div class='form-group'>
         <%= f.label :slug %>
-        <%= f.text_field :slug, class: 'input-md-12 form-control' %>
+        <%= f.text_field :slug, class: 'input-md-12 form-control', value: guide.slug || "/service-manual/" %>
```
tadast: Should we write a feature test to make sure the slug field is pre-filled when creating new guides?

I think I'd also prefer setting this value in the controller `new` action, the same way we set an edition for a new guide https://github.com/alphagov/service-manual-publisher/blob/616661944759aab8563bca2c3ede370c1a732e39/app/controllers/guides_controller.rb#L7

tadast:
LGTM :+1:

PR #23, opened 2015-10-21
Record the author of each guide edition

tadast:
Assign current user id as user_id on an edition. This is in order to keep track
of who's made changes to a guide over time. The latest editor is shown in the
list table of all guides.

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -1,6 +1,6 @@
 class GuidesController < ApplicationController
   def index
-    @guides = Guide.includes(:latest_edition)
+    @guides = Guide.includes(latest_edition: :user)
   end
 
   def new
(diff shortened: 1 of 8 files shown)


app/controllers/guides_controller.rb
```diff
@@ -35,10 +35,11 @@ def update
 private
 
   def guide_params
+    params[:guide][:latest_edition_attributes].merge!(user_id: current_user.id)
     params.require(:guide).permit(
       :slug,
       latest_edition_attributes: [
-        :title, :body, :description, :publisher_title, :phase,
+        :title, :body, :description, :publisher_title, :phase, :user_id,
         :related_discussion_href, :related_discussion_title, :update_type
```
tadast: I'm not a big fan of changes in this file, any better alternatives?

app/helpers/guide_helper.rb
```diff
@@ -6,4 +6,8 @@ def edit_action_label(guide)
       'Create new edition'
     end
   end
+
+  def last_editor_name(guide)
```
AndrewVos: `most_recent_editor_name` ?

AndrewVos: Or `latest_editor_name` ?

AndrewVos:
Is :user validated in the model?

tadast:
Thanks. Renamed the helper method and added a validation for user presence on Edition.

AndrewVos:
lgtm

PR #25, opened 2015-10-22
Display validation errors for guides

AndrewVos:
Show users what went wrong when they attempt to create or edit a guide
page.
Attempt to follow the guidance here: http://govuk-elements.herokuapp.com/errors/

This is a conversation starter for now. I'm interested in hearing how @tadast
thinks we could refactor this.

For now I have removed the anchor links to inputs, because it's quite hard to do in rails.

Whoever merges this branch should rebase the commits into one.

### DIFF
--- a/app/models/guide.rb
+++ b/app/models/guide.rb
@@ -3,7 +3,6 @@ require "gds_api/publishing_api"
 class Guide < ActiveRecord::Base
   validates :content_id, presence: true, uniqueness: true
   validates :slug, presence: true
-  validates_associated :latest_edition
   validates :slug, format: {
     with: /\A\/service-manual\//,
     message: "must be be prefixed with /service-manual/"
--- a/app/views/guides/_form.html.erb
+++ b/app/views/guides/_form.html.erb
@@ -3,7 +3,8 @@
     <fieldset class='meta'>
       <legend>Meta</legend>
       <div class='form-group'>
-        <%= f.label :slug %>
+        <%= f.label :slug  %>
+        <%= f.error_list :slug  %>
         <%= f.text_field :slug, class: 'input-md-12 form-control' %>
       </div>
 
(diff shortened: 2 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -1,2 +1,13 @@
...
+    [
+      object.class.name,
+      field,
+    ].map do |s|
+      s.to_s
+       .gsub(" ", "_")
+       .gsub(/\W/, '')
+       .downcase
+    end.join("_")
+  end
```
tadast: Would be nice to find a way to reuse the original logic that generates HTML `id`s 
https://github.com/rails/rails/blob/7f18ea14c893cb5c9f04d4fda9661126758332b5/actionview/lib/action_view/helpers/tags/base.rb#L99-L106

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,23 @@
 module ApplicationHelper
 end
+
+class ActionView::Helpers::FormBuilder
```
tadast: Should we move this to its own file somewhere under `lib`?

app/views/guides/_form.html.erb
```diff
@@ -3,34 +3,40 @@
     <fieldset class='meta'>
       <legend>Meta</legend>
       <div class='form-group'>
-        <%= f.label :slug %>
+        <%= f.label(:slug) %>
```
tadast: I think the convention is to omit parenthesis on method calls in views unless they're really needed.

app/helpers/application_helper.rb
AndrewVos: Absolutely. What about something like `app/builders`?

app/models/guide.rb
```diff
@@ -3,7 +3,6 @@
 class Guide < ActiveRecord::Base
   validates :content_id, presence: true, uniqueness: true
   validates :slug, presence: true
-  validates_associated :latest_edition
```
tadast: Minor: we spoke about it offline, but it would be nice to notice why you've removed this in the commit note.

app/helpers/application_helper.rb
tadast: That sounds alright, what about `{lib,app}/patches`, `{lib,app}/extensions`, `{lib,app}/enhancements` to emphasise the fact you're extending a standard class/module?

AndrewVos: I'm thinking `app/helpers/error_list_helper.rb`

tadast: That sounds appropriate for the current state of it. If anyone extends it further, they should rename the file imho.

PR #26, opened 2015-10-23
Granular versioning for Guides

tadast:
Record each draft save/publication as a separate edition.

Whenever changes are made, create a new guide edition. We used to do that for
currently published guides, but not for guides in draft state. Recording each
edition save aligns us with the Performance Platform team (responsible for
publishing-api) future plans for versioning. It also gives us better
accountability tracking as author of each change is now recorder.

Note that I opted to "clone" the previous edition before applying attributes
sent from the form. This is to ensure that we will not loose any data if it is
not passed from the form due to any potential changes in future.

This PR does not deal with versioning and optimistic-locking as the publishing-api changes to deal with that are still in development. We'll need to update this process a bit once the mentioned changes go live.

This PR also adds a minimalistic view for Guide history with aim to make it easier to demo, develop and design upcoming features.

### DIFF
--- a/app/controllers/editions_controller.rb
+++ b/app/controllers/editions_controller.rb
@@ -1,0 +1,6 @@
+class EditionsController < ApplicationController
+  def index
+    @guide = Guide.find(params[:guide_id])
+    @editions = @guide.editions.includes(:user).order(updated_at: :desc)
+  end
+end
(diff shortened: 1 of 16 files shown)


AndrewVos:
lgtm

tadast:
Thanks for looking at it :)

PR #27, opened 2015-10-27
Migrate to publishing-api V2

tadast:
The V2 version of publishing-api adds separate endpoints for tagging. We'll need to use them for creating Service Manual Categories ("Topics"). In preparation I've migrated the current process to publishing-api V2 without modifying behaviour.

However, the V2 has a significant difference in the way content gets published: the `publish` endpoint only accepts content_id and publishes the latest version of the document stored via `put_content`. I hesitated modifying the process we currently have in place knowing that @AndrewVos is currently working on changing it (you might want to rebase once this gets merged, Andrew).


AndrewVos:
lgtm

PR #28, opened 2015-10-27
Fix flash message styling

tadast:
The old markup was copied from another application and didn't really work with
the bootstrap version we're using. It also had extra marlup we're not using.


AndrewVos:
lg

PR #29, opened 2015-10-27
Allow guide pages to be reviewed by peers

AndrewVos:
Allow users to create Review Requests
Only allow guides to be published if they have been approved

### DIFF
--- a/app/controllers/review_requests_controller.rb
+++ b/app/controllers/review_requests_controller.rb
@@ -3,11 +3,9 @@ class ReviewRequestsController < ApplicationController
     review_request = ReviewRequest.create!
 
     guide = Guide.find(params[:review_request][:guide_id])
-    editions = guide.editions.where(review_request: nil)
-    editions.each do |edition|
-      edition.review_request = review_request
-      edition.save!
-    end
+    guide.editions
+      .where(review_request: nil)
+      .update_all(review_request_id: review_request)
 
     redirect_to root_path, notice: "Your review request has been created!"
   end
--- a/app/controllers/review_requests_controller.rb
+++ b/app/controllers/review_requests_controller.rb
@@ -1,0 +1,14 @@
+class ReviewRequestsController < ApplicationController
+  def create
+    review_request = ReviewRequest.create!
+
+    guide = Guide.find(params[:review_request][:guide_id])
+    editions = guide.editions.where(review_request: nil)
+    editions.each do |edition|
+      edition.review_request = review_request
+      edition.save!
+    end
+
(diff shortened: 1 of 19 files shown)


app/controllers/review_requests_controller.rb
```diff
@@ -0,0 +1,14 @@
+class ReviewRequestsController < ApplicationController
+  def create
+    review_request = ReviewRequest.create!
+
+    guide = Guide.find(params[:review_request][:guide_id])
+    editions = guide.editions.where(review_request: nil)
+    editions.each do |edition|
+      edition.review_request = review_request
+      edition.save!
+    end
```
tadast: I think it should be ok to do
```
guide.editions.where(review_request: nil).update_all(review_request: review_request)
```

app/models/guide.rb
```diff
@@ -6,6 +6,16 @@ class Guide < ActiveRecord::Base
     message: "must be be prefixed with /service-manual/"
   }
 
+  def needs_approval?
+    latest_edition.nil? ||
+    latest_edition.review_request.nil? ||
+      latest_edition.review_request.approvals.empty?
+  end
+
+  def needs_review?
```
tadast: Does it still need a review if it has an existing approval?

app/views/guides/index.html.erb
```diff
@@ -22,6 +23,14 @@
         <td class='last-edited-by'>
           <%= latest_editor_name(guide) %>
         </td>
+        <td>
+          <% if guide.latest_edition.review_request.present? && guide.latest_edition.review_request.approvals.any? %>
```
tadast: Would it make sense to encapsulate this logic as `Guide#current_draft_approvals` or something similar?
Looks a bit heavy, especially for a view.

spec/features/guide_publishing_flow_spec.rb
```diff
@@ -61,12 +61,59 @@
...
+  end
+
 private
 
-  def given_a_guide_exists(state:)
+  def given_a_guide_exists(state:, user: nil)
     edition = Generators.valid_edition
     edition.state = state
     edition.title = 'Sample Published Edition'
+    edition.user = user unless user.nil?
```
tadast: If I pass in `user: nil` to this method I would expect the guide edition to have no user, this code keeps the user assigned by the Generators. I think this is rather confusing.

tadast:
What happens to newly created editions after a review request is created? Do they get assigned the latest review_id until the draft is published?

tadast:
It would be nice to handle failure scenarios in the newly introduced controller actions.

PR #32, opened 2015-10-29
Review Requests

AndrewVos:
Users can request a review on guide editions. When a review is requested, a user can mark the edition as approved. Only then can the edition be published.

This stops users publishing content without a second pair of eyes looking over it.

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -14,7 +14,7 @@ class Edition < ActiveRecord::Base
   scope :review_requested, -> { where(state: 'review_requested') }
 
   validates_presence_of [:state, :phase, :description, :title, :update_type, :body, :publisher_title, :publisher_href, :user]
-  validates_inclusion_of :state, in: %w(draft published review_requested)
+  validates_inclusion_of :state, in: %w(draft published review_requested approved)
   validate :has_been_approved?
 
   def has_been_approved?
@@ -25,12 +25,6 @@ class Edition < ActiveRecord::Base
   end
 
   before_validation :assign_publisher_href
-
-  def ready_to_publish?
-    review_requested? &&
-      approvals.any? &&
-      !published?
-  end
 
   def draft?
     state == 'draft'
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -14,6 +14,14 @@ class Edition < ActiveRecord::Base
 
   validates_presence_of [:state, :phase, :description, :title, :update_type, :body, :publisher_title, :publisher_href, :user]
   validates_inclusion_of :state, in: %w(draft published)
+  validate :has_been_approved?
+
+  def has_been_approved?
+    return unless published?
+    if review_request.nil? || review_request.approvals.empty?
+      errors.add(:state, "This guide must be approved at least once before publishing")
+    end
+  end
 
   before_validation :assign_publisher_href
 
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -26,6 +26,12 @@ class Edition < ActiveRecord::Base
 
   before_validation :assign_publisher_href
 
+  def ready_to_publish?
+    review_requested? &&
+      approvals.any? &&
+      !published?
+  end
+
   def draft?
     state == 'draft'
   end
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -7,6 +7,8 @@ class Edition < ActiveRecord::Base
   belongs_to :guide
   belongs_to :user
 
+  belongs_to :review_request
+
   scope :draft, -> { where(state: 'draft') }
   scope :published, -> { where(state: 'published') }
 
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -7,18 +7,19 @@ class Edition < ActiveRecord::Base
   belongs_to :guide
   belongs_to :user
 
-  belongs_to :review_request
+  has_many :approvals
 
   scope :draft, -> { where(state: 'draft') }
   scope :published, -> { where(state: 'published') }
+  scope :review_requested, -> { where(state: 'review_requested') }
 
   validates_presence_of [:state, :phase, :description, :title, :update_type, :body, :publisher_title, :publisher_href, :user]
-  validates_inclusion_of :state, in: %w(draft published)
+  validates_inclusion_of :state, in: %w(draft published review_requested)
   validate :has_been_approved?
 
   def has_been_approved?
-    return unless published?
-    if review_request.nil? || review_request.approvals.empty?
+    return if !published?
+    if self.approvals.empty?
       errors.add(:state, "This guide must be approved at least once before publishing")
     end
   end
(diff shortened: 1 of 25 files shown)


app/models/edition.rb
```diff
@@ -7,11 +7,22 @@ class Edition < ActiveRecord::Base
...
+  scope :review_requested, -> { where(state: 'review_requested') }
 
   validates_presence_of [:state, :phase, :description, :title, :update_type, :body, :publisher_title, :publisher_href, :user]
-  validates_inclusion_of :state, in: %w(draft published)
+  validates_inclusion_of :state, in: %w(draft published review_requested approved)
+  validate :has_been_approved?
+
+  def has_been_approved?
+    return if !published?
+    if self.approvals.empty?
```
tadast: Minor: No need for `self` here

app/models/edition.rb
```diff
@@ -7,11 +7,22 @@ class Edition < ActiveRecord::Base
...
+      errors.add(:state, "This guide must be approved at least once before publishing")
```
tadast: I think it would be better to say: "All changes need to be approved by another Content Designer before publishing". it's the changes they're approving, not the guide itself.

app/views/guides/_form.html.erb
```diff
@@ -72,8 +72,29 @@
       <div class='form-group'>
         <%= f.submit "Save Draft", class: 'btn btn-success', name: :save_draft %>
         <%= link_to "Preview", document_preview_url(guide), class: 'btn btn-default' %>
-        <%= f.submit "Publish", class: 'btn btn-success', name: :publish %>
+        <% if guide.latest_edition.present? && guide.latest_edition.approved? %>
```
tadast: encapsulate this in `guide#ready_for_publishing?` ?

app/views/guides/_form.html.erb
```diff
@@ -72,8 +72,29 @@
...
+          <%= f.submit "Publish", class: 'btn btn-success', name: :publish %>
+        <% end %>
       </div>
     </fieldset>
   <% end %>
+
+  <% if guide.latest_edition.present? %>
+    <% if guide.latest_edition.persisted? && !guide.latest_edition.review_requested? %>
```
tadast: What happens if someone requests a review and then creates another edition? I think users only need to request a review once per-draft. What do you think?

app/views/guides/_form.html.erb
```diff
@@ -72,8 +72,29 @@
...
+      <%= form_for :review_request, url: edition_review_requests_path(edition_id: guide.latest_edition.id) do |f| %>
+        <%= f.submit "Send for review", class: 'btn btn-success' %>
+      <% end %>
+    <% end %>
+
+    <% if guide.latest_edition.persisted? && guide.latest_edition.review_requested? %>
```
tadast: Same here, i think we should check if there is any review request created since the last publication instead of checking the latest edition only.

spec/features/guide_publishing_flow_spec.rb
```diff
@@ -61,6 +61,42 @@
...
 
+  context "with a review requested" do
+    it "lists editions that need a review" do
+      edition = Generators.valid_edition(approvals: [])
+      guide = Guide.create!(latest_edition: edition, slug: "/service-manual/something")
+
+      visit edit_guide_path(guide)
+      click_button "Send for review"
+      visit guides_path
+      expect(page).to have_content "review_requested"
```
tadast: Should we `titleize` this value?

spec/features/guide_publishing_flow_spec.rb
```diff
@@ -61,6 +61,42 @@
...
+
+    context "approved by another user" do
+      it "lists editions that are approved" do
+        edition = Generators.valid_edition(state: "review_requested", approvals: [])
+        guide = Guide.create!(latest_edition: edition, slug: "/service-manual/something")
+
+        reviewer = User.new(name: "Some User")
+        login_as reviewer
+        visit guides_path
+        click_link "Create new edition"
```
tadast: That's not a great UX I think, perhaps we could make the "state" link clickable?

spec/features/guide_publishing_flow_spec.rb
```diff
@@ -61,6 +61,42 @@
...
+        click_button "Mark as Approved"
+        expect(page).to have_content "Thanks for approving this guide"
+        expect(page).to have_content "approved"
+      end
```
tadast: I think we should have another test that specifically tests that the original author can not approve their own review request.

spec/features/guide_spec.rb
```diff
@@ -60,18 +60,29 @@
...
     expect(api_double).to receive(:put_content)
-                            .twice
+                            .thrice
                             .with(an_instance_of(String), be_valid_against_schema('service_manual_guide'))
     expect(api_double).to receive(:publish)
                             .twice
                             .with(an_instance_of(String), 'minor')
 
+    click_button "Save Draft"
+    guide = Guide.first
```
tadast: Minor: this feels a bit brittle, I usually do something more specific to ensure the right record is returned, e.g. `Guide.find_by_slug`

spec/models/edition_spec.rb
```diff
@@ -19,15 +19,40 @@
     end
 
     it "allows 'published' state" do
-      edition = Edition.new(state: 'published')
+      edition = Edition.new(
+        state: 'published',
+        approvals: [Approval.new(user:User.first)],
```
tadast: Weird spacing here `user: User.first` :)

spec/models/edition_spec.rb
```diff
@@ -19,15 +19,40 @@
...
+      )
+      edition.valid?
+
+      expect(edition.errors.full_messages_for(:state).size).to eq 0
+    end
+
+    it "allows 'review_requested' state" do
+      edition = Edition.new(
+        state: 'review_requested',
+        approvals: [Approval.new(user:User.first)],
```
tadast: Why do we need an approval for this state?

spec/models/edition_spec.rb
```diff
@@ -19,15 +19,40 @@
...
       edition.valid?
 
       expect(edition.errors.full_messages_for(:state).size).to eq 0
     end
 
     it "does not allow arbitrary values" do
-      edition = Edition.new(state: 'supercharged')
+      edition = Edition.new(
+        state: 'supercharged',
+        approvals: [Approval.new(user:User.first)],
```
tadast: Do we need an approval for this test?

spec/support/generators.rb
```diff
@@ -8,7 +10,8 @@ def self.valid_edition(attributes = {})
       update_type:     "major",
       body:            "# Heading",
       publisher_title: Edition::PUBLISHERS.keys.first,
-      user:            User.new(name: "Generated User")
+      user:            user,
+      approvals:       [Approval.new(user: user)],
```
tadast: I think we should not add any values to this generator unless they are absolutely required to pass validations. This will make it very confusing long term i.e. as a new developer I wouldn't expect a `valid_edition` to have any associated approvals. It also makes test setup slower globally

AndrewVos: Yeah, but it's a _valid_ edition, which means it validates.

spec/models/edition_spec.rb
AndrewVos: We do, but if we use Generator then no.

tadast: Sorry, I'm not following: we're testing errors on the state field, how is that field affected by the presence of approvals?

PR #33, opened 2015-10-29
Review Requests

AndrewVos:
Users can request a review on guide editions. When a review is requested, a user can mark the edition as approved. Only then can the edition be published.

This stops users publishing content without a second pair of eyes looking over it.

At the moment we don't ensure that a user cannot approve their own
edition. This is to allow easier UI testing. There is a pending test for
it, so we don't forget.

### DIFF
--- a/app/helpers/guide_helper.rb
+++ b/app/helpers/guide_helper.rb
@@ -7,6 +7,21 @@ module GuideHelper
     end
   end
 
+  STATE_CSS_CLASSES = {
+    "new"              => "default",
+    "draft"            => "danger",
+    "review_requested" => "warning",
+    "approved"         => "success",
+    "published"        => "info",
+  }
+
+  def state_label(guide)
+    state     = guide.latest_edition.try(:state) || "new"
+    title     = state.titleize
+    css_class = STATE_CSS_CLASSES[state]
+    content_tag :span, title, title: 'State', class: "label label-#{css_class}"
+  end
+
   def latest_editor_name(guide)
     guide.latest_edition.user.try(:name).to_s
   end
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -7,12 +7,14 @@ class Edition < ActiveRecord::Base
   belongs_to :guide
   belongs_to :user
 
+  has_many :approvals
+
   scope :draft, -> { where(state: 'draft') }
   scope :published, -> { where(state: 'published') }
   scope :review_requested, -> { where(state: 'review_requested') }
 
   validates_presence_of [:state, :phase, :description, :title, :update_type, :body, :publisher_title, :publisher_href, :user]
-  validates_inclusion_of :state, in: %w(draft published review_requested)
+  validates_inclusion_of :state, in: %w(draft published review_requested approved)
 
   before_validation :assign_publisher_href
 
@@ -26,6 +28,10 @@ class Edition < ActiveRecord::Base
 
   def review_requested?
     state == 'review_requested'
+  end
+
+  def approved?
+    state == 'approved'
   end
 
   def copyable_attributes(extra_attributes = {})
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -9,9 +9,10 @@ class Edition < ActiveRecord::Base
 
   scope :draft, -> { where(state: 'draft') }
   scope :published, -> { where(state: 'published') }
+  scope :review_requested, -> { where(state: 'review_requested') }
 
   validates_presence_of [:state, :phase, :description, :title, :update_type, :body, :publisher_title, :publisher_href, :user]
-  validates_inclusion_of :state, in: %w(draft published)
+  validates_inclusion_of :state, in: %w(draft published review_requested)
 
   before_validation :assign_publisher_href
 
@@ -21,6 +22,10 @@ class Edition < ActiveRecord::Base
 
   def published?
     state == 'published'
+  end
+
+  def review_requested?
+    state == 'review_requested'
   end
 
   def copyable_attributes(extra_attributes = {})
--- a/app/models/guide.rb
+++ b/app/models/guide.rb
@@ -16,7 +16,7 @@ class Guide < ActiveRecord::Base
   end
 
   def work_in_progress_edition?
-    latest_edition.try(:state) == 'draft'
+    latest_edition.try(:published?) == false
   end
 
   def can_request_review?
@@ -24,6 +24,13 @@ class Guide < ActiveRecord::Base
     return false if !latest_edition.persisted?
     return false if latest_edition.review_requested?
     return false if latest_edition.published?
+    return false if latest_edition.approved?
     true
+  end
+
+  def can_mark_as_approved?
+    return false if latest_edition.nil?
+    return false if !latest_edition.persisted?
+    latest_edition.review_requested?
   end
 end
--- a/app/models/guide.rb
+++ b/app/models/guide.rb
@@ -18,4 +18,12 @@ class Guide < ActiveRecord::Base
   def work_in_progress_edition?
     latest_edition.try(:state) == 'draft'
   end
+
+  def can_request_review?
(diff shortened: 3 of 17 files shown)


app/models/edition.rb
```diff
@@ -7,11 +7,22 @@ class Edition < ActiveRecord::Base
...
+  scope :review_requested, -> { where(state: 'review_requested') }
 
   validates_presence_of [:state, :phase, :description, :title, :update_type, :body, :publisher_title, :publisher_href, :user]
-  validates_inclusion_of :state, in: %w(draft published)
+  validates_inclusion_of :state, in: %w(draft published review_requested approved)
+  validate :has_been_approved?
+
+  def has_been_approved?
+    return if !published?
+    if self.approvals.empty?
```
tadast: no need for "self"

app/views/guides/_form.html.erb
```diff
@@ -72,8 +72,29 @@
...
-        <%= f.submit "Publish", class: 'btn btn-success', name: :publish %>
+        <% if guide.latest_edition.present? && guide.latest_edition.approved? %>
+          <%= f.submit "Publish", class: 'btn btn-success', name: :publish %>
+        <% end %>
       </div>
     </fieldset>
   <% end %>
+
+  <% if guide.latest_edition.present? %>
+    <% if guide.latest_edition.persisted? && !guide.latest_edition.review_requested? %>
```
tadast: Extract method?

app/views/guides/_form.html.erb
```diff
@@ -72,8 +72,29 @@
...
+      <%= form_for :review_request, url: edition_review_requests_path(edition_id: guide.latest_edition.id) do |f| %>
+        <%= f.submit "Send for review", class: 'btn btn-success' %>
+      <% end %>
+    <% end %>
+
+    <% if guide.latest_edition.persisted? && guide.latest_edition.review_requested? %>
```
tadast: Extract method?

app/views/guides/index.html.erb
```diff
@@ -22,6 +19,9 @@
         <td class='last-edited-by'>
           <%= latest_editor_name(guide) %>
         </td>
+        <td>
+          <%= guide.latest_edition.state %>
```
tadast: `...state.titleize`?

spec/features/guide_publishing_flow_spec.rb
```diff
@@ -61,12 +63,56 @@
...
+        edition = Generators.valid_edition(state: "review_requested", approvals: [])
+        guide = Guide.create!(latest_edition: edition, slug: "/service-manual/something")
+
+        reviewer = User.new(name: "Some User")
+        login_as reviewer
+        visit guides_path
+        click_link "Create new edition"
+        click_button "Mark as Approved"
+        expect(page).to have_content "Thanks for approving this guide"
+        expect(page).to have_content "approved"
```
tadast: This got me thinking, should we show who are the reviewers somewhere on the page?

spec/features/guide_publishing_flow_spec.rb
```diff
@@ -61,12 +63,56 @@
...
+
+  context "without any approvals" do
+    it "does not allow guides to be published" do
+      edition = Generators.valid_edition(state: "review_requested", approvals: [])
+      guide = Guide.create!(latest_edition: edition, slug: "/service-manual/something")
+      visit edit_guide_path(guide)
+      expect(page).to_not have_button "Publish"
+    end
+  end
+
```
tadast: I think we should have a test to prove the original author can not approve their own changes

AndrewVos: I've purposely left this out for now, to allow easier testing in browser. Would you be happy if I added a pending test for now?

tadast: It's easy enough to forget, so yeah, a pending test would be ok. Ideally we'd have an environment variable that we could set in dev environment to relax the rules on who can approve

spec/features/guide_publishing_flow_spec.rb
AndrewVos: Yeah, I would like to show them on the edit_guide page.

app/models/edition.rb
```diff
@@ -7,11 +7,15 @@ class Edition < ActiveRecord::Base
...
+  has_many :approvals
+
   scope :draft, -> { where(state: 'draft') }
   scope :published, -> { where(state: 'published') }
+  scope :review_requested, -> { where(state: 'review_requested') }
 
   validates_presence_of [:state, :phase, :description, :title, :update_type, :body, :publisher_title, :publisher_href, :user]
-  validates_inclusion_of :state, in: %w(draft published)
+  validates_inclusion_of :state, in: %w(draft published review_requested approved)
+  validate :approved?, if: :review_requested?
```
tadast: What is this supposed to do?

app/models/guide.rb
```diff
@@ -18,4 +18,17 @@ class Guide < ActiveRecord::Base
   def work_in_progress_edition?
     latest_edition.try(:state) == 'draft'
   end
+
+  def can_request_review?
+    return false if latest_edition.nil?
+    return false if !latest_edition.persisted?
+    return false if latest_edition.review_requested?
```
tadast: What if the latest edition has been published, will it be covered by `latest_edition.review_requested?`

app/models/edition.rb
tadast: http://api.rubyonrails.org/classes/ActiveModel/Validations/ClassMethods.html#method-i-validate

spec/models/guide_spec.rb
```diff
@@ -11,11 +11,64 @@
...
       edition = Guide.new(slug: "/something", latest_edition: edition)
       edition.valid?
       expect(edition.errors.full_messages_for(:slug)).to eq ["Slug must be be prefixed with /service-manual/"]
     end
   end
 
+  context "review and approval" do
+    let :edition do
+      edition = Generators.valid_edition
+      allow(edition).to receive(:persisted?) { true }
```
tadast: you seem to be overriding this line in a few tests, perhaps you can move this stub to a before block in contexts where it's needed only

tadast:
Perhaps we should update https://github.com/alphagov/service-manual-publisher/blob/master/app/models/guide.rb#L19, as it now shows "Create new edition" for any edition that is not draft (i.e. requested review, approved etc.)

tadast:
If I approve changes, I can see "Send for Review" button again, If I click it, it looses approval even though there have been no changes. I think we should hide "Send for review" when all changes are approved.

KushalP:
I've looked at this pull request as individual commits and then looking at it as the final diff. The final piece of work looks good, but the some of the intermediate commits are a little confusing.

One suggestion I have for improving it is to rebase out the addition and removal of migrations and bits of code. It will be a lot easier for others to understand the progression.

app/helpers/guide_helper.rb
```diff
@@ -7,6 +7,21 @@ def edit_action_label(guide)
...
+    "review_requested" => "warning",
+    "approved"         => "success",
+    "published"        => "info",
+  }
+
+  def state_label(guide)
+    state     = guide.latest_edition.try(:state) || "new"
+    title     = state.titleize
+    css_class = STATE_CSS_CLASSES[state]
+    "<span class='label label-#{css_class}' title='State'>#{title}</span>".html_safe
```
tadast: ```content_tag :span, title, title: 'State', class: "label label-#{css_class}"```

spec/features/guide_publishing_flow_spec.rb
tadast: Perhaps it's worth mentioning in the commit note and PR description?

AndrewVos: Done

tadast:
I like the commit history after rebasing :100: :+1: LGTM

PR #34, opened 2015-10-30
Add pagination with kaminari

AndrewVos:
This uses the bootstrap3 theme from
https://github.com/amatsuda/kaminari_themes


Gemfile
```diff
@@ -13,6 +13,7 @@ gem 'govuk_admin_template', '~> 3.0.0'
 gem 'gds-sso', '~> 11.0.0'
 gem 'gds-api-adapters', '~> 24.6.0'
 gem 'govspeak', '~> 3.4.0'
+gem 'kaminari'
```
KushalP: Please can you set an explicit version number, as per the [styleguide](https://github.com/alphagov/styleguides/blob/master/ruby.md#gemfile).

AndrewVos: Done

PR #35, opened 2015-11-02
Improvements to previewing of drafts

tadast:
An issue has been reported by our team member who noticed that the "Preview" button is not working on new guides. I've researched this and uncovered some more problems, that I've written up [on trello](https://trello.com/c/dIVi531R/35-improve-draft-previewing-functionality) and in commit messages in this PR. 

In short: make sure no work is lost when previewing and content designers always see the latest draft version by saving all work before previewing.

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -19,7 +19,7 @@ class GuidesController < ApplicationController
     ActiveRecord::Base.transaction do
       if @guide.save
         GuidePublisher.new(guide: @guide).process
-        redirect_to root_path, notice: "Guide has been created"
+        redirect_to success_url(@guide), notice: "Guide has been created"
       else
         render action: :new
       end
@@ -43,7 +43,7 @@ class GuidesController < ApplicationController
     ActiveRecord::Base.transaction do
       if @guide.update_attributes(guide_params({latest_edition_attributes: {state: edition_state_from_params, user_id: current_user.id}}))
         GuidePublisher.new(guide: @guide).process
-        redirect_to root_path, notice: "Guide has been updated"
+        redirect_to success_url(@guide), notice: "Guide has been updated"
       else
         render action: :edit
       end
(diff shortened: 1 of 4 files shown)


AndrewVos:
After speaking to Tom Hughes, one thing that is really irritating is that the Preview button in publisher is not a link. This means it can't be opened in a new page.

I would suggest:
- Make the Preview button a link.
- Only show it when the model has been persisted.
- Do *not* save when hitting preview. This isn't normal behaviour and might be unexpected to content designers.

tadast:
I've checked with Tom and it looks his frustration is that the Preview button in mainstream publisher does not open the preview in a new page by default.

I've also asked a few more people from the Content team:

```
Tadas: Hello, imagine an ideal world where you are helping to design a new publishing tool. For previewing your work in progress draft, would you ideally have a button "Save Draft and Preview" or a "Preview" button that requires to save draft separately?
Abigail: Combined is good as long as it has a separate save draft button for when I didn't want to preview
Keith: 'Save and preview' makes sense, as long as there is a separate 'Save' like Abi mentioned
Lucy:  I like being able to save as I work on a draft and then preview once I've made a substantial amount of changes. I like to do these actions separately
Tadas: Lucy, if you have two buttons "Save Draft" and "Preview" and then make some changes in the form and click Preview (without clicking Save Draft), would you expect those changes to be shown on preview or would you expect to see the version of the draft from the last time you clicked Save Draft?
Lucy: This might not be useful, but it depends where the button was. If it was next to save draft, I would expect it to show me the content without my changes. If it was closer to content fields, I would expect it to show me the content with my changes. Does that make sense?
```

So far it sounds like being able to Save Draft without previewing is important. I still think having a Preview button that doesn't save changes would be ambiguous. Having it named "Save Draft and Preview" should remove this ambiguity. I'll ask Content Designers in our team what they think.

AndrewVos:
It might be ambigious, but remember the people using this would get used to a certain behaviour very quickly.

I don't think the preview button should save, for the simple reason that the verb "preview" does not imply "save".

tadast:
@AndrewVos I've modified the behaviour you were concerned about in the commit titled "Avoid weird errors when saving new guides". Checkout the branch and try it out, I think it works better than before now.

AndrewVos:
lgtm

PR #36, opened 2015-11-02
Disallow guide slug clashes

AndrewVos:
Wrap the saving of guides in a transaction, so that if publishing to the
draft content store fails, then we do not store an object in the
database.

### DIFF
--- a/spec/features/guide_spec.rb
+++ b/spec/features/guide_spec.rb
@@ -95,16 +95,66 @@ RSpec.describe "creating guides", type: :feature do
     expect(edition.published?).to eq true
   end
 
-  it "shows api errors" do
-    fill_in_guide_form
+  context "when creating a new guide" do
+    context "when publishing raises an exception" do
+      before do
+        api_error = GdsApi::HTTPClientError.new(422, "Error message stub", "error" => { "message" => "Error message stub" })
+        expect_any_instance_of(GdsApi::PublishingApiV2).to receive(:put_content).and_raise(api_error)
+      end
 
-    api_error = GdsApi::HTTPClientError.new(422, "Error message stub", "error" => { "message" => "Error message stub" })
-    expect_any_instance_of(GdsApi::PublishingApiV2).to receive(:put_content).and_raise(api_error)
+      it "shows api errors" do
+        fill_in_guide_form
+        click_button "Save Draft"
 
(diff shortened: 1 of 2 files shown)


tadast:
Cool! Do we need to do the same with the update action?

AndrewVos:
Spoke to @tadast offline, and yes it would be good to have the update action in a transaction too.

AndrewVos:
Ok done @tadast

spec/features/guide_spec.rb
```diff
@@ -95,16 +95,64 @@
...
+        expect(Guide.count).to eq 0
+        expect(Edition.count).to eq 0
+      end
+    end
+  end
 
-    api_error = GdsApi::HTTPClientError.new(422, "Error message stub", "error" => { "message" => "Error message stub" })
-    expect_any_instance_of(GdsApi::PublishingApiV2).to receive(:put_content).and_raise(api_error)
+  context "when updating a guide" do
+    context "with an api that fails the second time" do
```
tadast: I find this description a bit confusing. Do you mean "with an api that fails"?

spec/features/guide_spec.rb
```diff
@@ -95,16 +95,64 @@
...
+
+        visit edit_guide_path(guide)
+        click_button "Save Draft"
+
+        within ".alert" do
+          expect(page).to have_content('Error message stub')
+        end
+      end
+
+      it "does not store a guide" do
```
tadast: `it "does not save the guide and persist an extra edition" do`?

spec/features/guide_spec.rb
```diff
@@ -95,16 +95,64 @@
...
+        edition = Generators.valid_edition(title: "something")
+        guide = Guide.create!(slug: "/service-manual/something", latest_edition: edition)
+
+        expect_any_instance_of(GuidePublisher).to receive(:process).once.and_raise(api_error)
+
+        visit edit_guide_path(guide)
```
tadast: should we try to change a field, try saving and then assert that the field did not change?

AndrewVos: Sure

spec/features/guide_spec.rb
AndrewVos: Done

PR #37, opened 2015-11-02
Show the most recently edited Guides on top of the index page

tadast:
We used to rely on default Rails ordering which is ordering Guides based on
their ids. I believe it is more natural to show the most recently updated
Guides on top, because they are usually the ones requiring attention.


AndrewVos:
lgtm

PR #38, opened 2015-11-03
Allow users to comment on editions

AndrewVos:
At the moment, we only see comments for the latest edition. We might
want to change this at some point.

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -30,11 +30,19 @@ class GuidesController < ApplicationController
   def edit
     @guide = Guide.find(params[:id])
     @edition = @guide.latest_edition.unsaved_copy
+    @comments = @guide.latest_edition.comments
+      .order(created_at: :asc)
+      .includes(:user)
+    @new_comment = @guide.latest_edition.comments.build
   end
 
   def update
     @guide = Guide.find(params[:id])
     @edition = build_new_edition_version_for(@guide)
+    @comments = @guide.latest_edition.comments
+      .order(created_at: :asc)
+      .includes(:user)
+    @new_comment = @guide.latest_edition.comments.build
     ActiveRecord::Base.transaction do
       if @guide.update_attributes(guide_params)
         GuidePublisher.new(guide: @guide, edition: @edition).process
--- a/db/migrate/20151103120443_create_comments.rb
+++ b/db/migrate/20151103120443_create_comments.rb
@@ -1,0 +1,19 @@
+class CreateComments < ActiveRecord::Migration
+  def self.up
+    create_table :comments do |t|
+      t.text :comment
+      t.references :commentable, :polymorphic => true
+      t.references :user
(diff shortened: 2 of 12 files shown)


app/views/guides/_form.html.erb
```diff
@@ -77,3 +77,35 @@
     </fieldset>
   <% end %>
 </div>
+
+<div class='well comments'>
```
tadast: I don't think this should live in this file. Perhaps you can have it in edit.html.erb?

app/views/guides/_form.html.erb
```diff
@@ -77,3 +77,35 @@
...
+  <fieldset class='meta'>
+    <legend>Comments</legend>
+    <% if guide.latest_edition.present? %>
+      <% guide.latest_edition.comments.includes(:user).each do |comment| %>
```
tadast: I think it would be better to make this query in the controller and expose it as an instance variable @comments

db/migrate/20151103120443_create_comments.rb
```diff
@@ -0,0 +1,20 @@
+class CreateComments < ActiveRecord::Migration
+  def self.up
+    create_table :comments do |t|
+      t.string :title, :limit => 50, :default => ""
```
tadast: Do we need this?

spec/features/guide_spec.rb
```diff
@@ -108,6 +108,27 @@
...
   end
 
+  it "allows discourse" do
+    edition = Generators.valid_edition
+    guide = Guide.create!(
+      latest_edition: edition,
+      slug: "/service-manual/test/slug_published"
+    )
+
+    user = User.create!(name: "The user that comments")
```
tadast: I believe this user is not used for anything in this test

app/views/guides/_form.html.erb
AndrewVos: Why? It's doing the same sort of thing as everything else in this file?

tadast:
It currently shows comments on the latest edition only. Ideally it would show all comments since the latest published edition. If you have planned this for future then it's worth explaining this temporary behaviour in the PR description and commit note.

app/controllers/guides_controller.rb
```diff
@@ -27,6 +27,8 @@ def create
   def edit
     @guide = Guide.find(params[:id])
     @edition = @guide.latest_edition.unsaved_copy
+    @comments = @guide.latest_edition.comments.includes(:user)
+    @new_comment = @guide.latest_edition.comments.build
```
tadast: last one: can we get rid of default_scope and order comments in here instead? It's subjective, but I really dislike default_scope, it has always made me regret using it in the past :/

AndrewVos: Yeah, that was what the gem generated... Removed it now.

tadast:
LGTM :+1: needs rebasing on master though :)

PR #39, opened 2015-11-04
Only duplicate editions when creating a new draft

AndrewVos:
- Comments, and approvals, will stick around on the edition even when it is published.
- Only published versions of editions will stay around forever
- The code in GuidesController should be a lot easier to understand
- I have used inline styles for the "Create a new edition" form for now. Sorry.

### DIFF
--- a/app/views/guides/index.html.erb
+++ b/app/views/guides/index.html.erb
@@ -23,7 +23,13 @@
           <%= state_label(guide) %>
         </td>
         <td class='content-controls'>
-          <%= link_to edit_action_label(guide), edit_guide_path(guide), class: 'btn btn-default btn-xs' %>
+          <% if guide.latest_edition.published? %>
+            <%= form_for :drafts, url: guide_drafts_path(guide_id: guide.id), html: {style: "display: inline;"} do |f| %>
+              <%= f.submit("Create new edition", class: 'btn btn-default btn-xs') %>
+            <% end %>
+          <% else %>
+            <%= link_to "Continue editing", edit_guide_path(guide), class: 'btn btn-default btn-xs' %>
+          <% end %>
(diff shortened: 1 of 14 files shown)


tadast:
It would be nice to explain why we are going back to this behaviour and what are the implications of this change in the PR description.

AndrewVos:
@tadast ready for rereview

app/views/guides/index.html.erb
```diff
@@ -23,7 +23,13 @@
           <%= state_label(guide) %>
         </td>
         <td class='content-controls'>
-          <%= link_to edit_action_label(guide), edit_guide_path(guide), class: 'btn btn-default btn-xs' %>
+          <% if guide.latest_edition.published? %>
+            <%= form_for :drafts, url: guide_drafts_path(guide_id: guide.id), html: {style: "display: inline;"} do |f| %>
```
tadast: :laughing: I guess inline styles will go away when we have the actual designs :)

AndrewVos: Yes they will!

spec/features/guide_publishing_flow_spec.rb
```diff
@@ -7,59 +7,79 @@
...
-  it "should create a new edition even when saving a draft" do
-    guide = given_a_guide_exists state: 'draft'
-    visit guides_path
-    link = there_should_be_a_control_link "Continue editing", document: guide
-    link.click
-    fill_in "Title", with: "Sample Published Edition 2"
-    click_button "Save Draft"
-    expect(current_path).to eq root_path
+  context "latest edition is not published" do
+    it "should not create a new draft edition" do
```
tadast: "should not save an extra draft if someone else clicks the link in the meantime" ?

AndrewVos: Done

PR #40, opened 2015-11-04
Spike: Service Manual Section page hierarchy

tadast:
:warning: Don't merge, this is for discussion only

Since we can't use any existing publishers for Service Manual Section pages ([an example of a section page](http://sm-11.herokuapp.com/agile-delivery/)) and don't want to build our own just yet, I suggest we use a rake task that publishes a hardcoded Service Manual Section payloads. This spike is intended as a proof of concept. There are two YAML files produced by scraping http://sm-11.herokuapp.com: one with a list of Guide payloads, and one with a list of Section page payloads. 

Interesting points:

- It does not have any data that deals with breadcrumbs, publishing platform
  team is working on it.
- It is using links hash with content IDs alongside `details -> link_groups  -> linked items`,
  the former allows the front-end to get all required info about linked items (title, href) and the later is used for grouping the linked items.
- Guides are scraped just to have some real content to link to from the Section pages: guides are not persisted locally, their bodies are HTML, not GOVSPEAK, there are no sidebar links etc but it is all irrelevant as we'll have real Guide pages saved when we do the real thing :)
- It posts links alongside content. In v2 APIs linking/tagging is a separate step so the process in real application would involve more steps.

The remaining part of this spike that renders this data in the front end is here:
https://github.com/alphagov/government-frontend/compare/service_manual_section_spike

## How to try this out locally

```
cd government-frontend
git checkout service_manual_section_spike
cd service-manual-publisher
git pull --rebase
git checkout section_hierarchy_spike
# run/restart both apps
rake hierarchy_spike:publish_guides
rake hierarchy_spike:publish_sections
```

Now you should be able to visit section and guide pages that just got published. Their base paths correspond to the ones in http://sm-11.herokuapp.com, but start with http://government-frontend.dev.gov.uk/service-manual i.e. to see [Agile Delivery](sm-11.herokuapp.com/agile-delivery) equivalent, visit http://government-frontend.dev.gov.uk/service-manual/agile-delivery

## The way this could work

Once we have a section with a few content items we're happy to publish in the publisher app, we could create a rake task similar to the one proposed here that would know which persisted content guides fall under which section. Based on hardcoded base_paths or titles it would fetch content IDs from either a local database or content store and compile a Service Manual Section payload that could then be published via the publishing-api. 

In order to make changes in the hierarchy we'd need to update the rake task in git, deploy and re-run it, but it should be ok for initial stages.

### DIFF
--- a/lib/tasks/hierarchy_spike.rake
+++ b/lib/tasks/hierarchy_spike.rake
@@ -25,6 +25,7 @@ namespace :hierarchy_spike do
     with_payload(file_name) do |payload|
       validate_against_schema(payload)
       publishing_api.put_content(payload[:content_id], payload)
+      publishing_api.put_links(payload[:content_id], payload) if payload[:links].present?
       publishing_api.publish(payload[:content_id], 'minor')
     end
   end
--- a/lib/tasks/hierarchy_spike.rake
+++ b/lib/tasks/hierarchy_spike.rake
@@ -1,0 +1,34 @@
+require 'yaml'
+require "gds_api/publishing_api_v2"
+
+# The YAML data files are generated using a script
+# https://gist.github.com/tadast/defd1b6fc1e2eccd77f3
+
+namespace :hierarchy_spike do
+  desc "Store hard-coded Service Manual Section hierarchy to the publishing API as a draft"
+  task publish_guides: [:environment] do
+    publish_from_file("guides.yml")
+  end
+
+  desc "Publish hard-coded Service Manual Section hierarchy that was previously saved as a draft"
+  task publish_sections: [:environment] do
+    publish_from_file("section_hierarchy.yml")
+  end
+
+  def publish_from_file(file_name)
+    publishing_api = GdsApi::PublishingApiV2.new(Plek.new.find('publishing-api'))
+    with_payload(file_name) do |payload|
+      publishing_api.put_content(payload[:content_id], payload)
+      publishing_api.publish(payload[:content_id], 'minor')
+    end
+  end
+
+  def with_payload(file_name)
+    payloads = YAML.load_file(Rails.root.join('lib', 'tasks_data', file_name))
+    payloads.each do |payload|
+      print "Publishing #{payload[:title]} at #{payload[:base_path]}"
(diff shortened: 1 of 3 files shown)


tadast:
From a slack discussion:

Perhaps in production a realistic sequence of events would be:

- A few guide pages are created without links
- A Service Manual Section page is created without links nor grouped links
- The guide pages are tagged to belong to the section page
- In whatever application is publishing section pages linked guides are curated to linked groups
- Another API call would need to be made to update Section page links from the linked groups :question:

tadast:
Elliot on slack about why links should be posted separately from content in v2:

> links are versioned separately to content, and so there's no way to do optimistic locking for the links if you're sending both to the content endpoint
> because that version only refers to the content itself

lib/tasks/hierarchy_spike.rake
```diff
@@ -0,0 +1,48 @@
...
+  task publish_sections: [:environment] do
+    publish_from_file("section_hierarchy.yml")
+  end
+
+  def publish_from_file(file_name)
+    publishing_api = GdsApi::PublishingApiV2.new(Plek.new.find('publishing-api'))
+    with_payload(file_name) do |payload|
+      validate_against_schema(payload)
+      publishing_api.put_content(payload[:content_id], payload)
+      publishing_api.put_links(payload[:content_id], payload) if payload[:links].present?
```
tadast: I just learned from @elliotcm that putting links now takes immediate effect in production, therefore this leaves potential for data in `links` and `link_groups -> linked_items` to get out of sync as this operation is not atomic. Even worse, once we

Moreover, once we have a [UI for building section pages](https://github.com/alphagov/service-manual-publisher/pull/40#issuecomment-154373882) we'd need to work around this by making sure we always update the links hash only and every time time we publish content. @KushalP any thoughts?

KushalP: This feels like something we should talk to the publishing platform team about. Ideally it would be something that they'd take into account and we'd act on accordingly.

tadast:
Closing in favour of #62

PR #41, opened 2015-11-06
Editors can leave a change note on editions

AndrewVos:
```
As someone working on a government service
I want to know what has changed and why
So that I can see if it is relevant to me/understand more about I need to do
```

Provide a summary of what has changed and why.

- If a user selects a 'major change' then a change note field is shown. (either on a new edition or a new draft).
- Change notes are a mandatory field.
- New drafts have change notes

Also:

- Move all edition fields into one #fields_for (was getting a bit
unwieldy).
- Stop using `fields_for :latest_edition`, instead use `fields_for :editions` because latest_edition was acting inconsistently with the has_one association.
- Remove the "Republish" and empty options from `update_type`, as
  they're not in the spec.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,0 +1,15 @@
+// This is a manifest file that'll be compiled into application.js, which will include all the files
+// listed below.
+//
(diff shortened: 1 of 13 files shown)


tadast:
There are some bits in this PR that are not ideal but are going to change soon as I'm working on another PR, so it LGTM :+1:

PR #43, opened 2015-11-09
Use /tmp for db:seed script

AndrewVos:
We're going to be running this script in a preview environment, so it
would be safer if we only used tmp directories. We don't want to
overwrite anything.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,12 +1,14 @@
 if Rails.env.development?
-  unless Dir.exist?("../government-service-design-manual")
+  directory = File.join(Dir.mktmpdir("government-service-design-manual"), "git")
+  unless Dir.exist?(directory)
     print "Cloning repository..."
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -1,12 +1,14 @@
 if Rails.env.development?
```
tadast: We need to change the line above as the preview servers run in production env. Not sure if we need to add an additional ENV variable to prevent this from being run on production servers. 

Alternatively we could exit `if Rails.env.production? && Guide.any?`

KushalP: > We need to change the line above as the preview servers run in production env. Not sure if we need to add an additional ENV variable to prevent this from being run on production servers.

I'd rather that be something we run/set manually for this task. We don't want someone to run it accidentally in production.

tadast: You mean something like `I_UNDERSTAND_THIS_IS_GOING_TO_ADD_SEED_DATA_TO_PRODUCTION_DB=true rake db:seed`?

PR #44, opened 2015-11-09
Allow running db:seed on preview server

AndrewVos:
Test GOVUK_APP_DOMAIN env variable to see if the script is running on
preview.


tadast:
LGTM

PR #45, opened 2015-11-10
Set phase for guide editions to 'alpha' by default

tadast:
This field was first introduced because it's a required field by default in
govuk-content-schemas. It has now become clear that we'll only use it to
indicate to the front-end application whether it should display an alpha/beta
banner. The banner would always be the same across all Service Manual content
so there is no need to display it in the UI. I've set it to 'alpha' initially,
once we decide to migrate to beta we'll need to do it with all persisted guides
locally as well as migrating published guides via publishing-api.

As a part of this effort https://github.com/alphagov/government-frontend/pull/65 will need to be updated to respect the `phase` field value.

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -85,7 +85,6 @@ private
         :body,
         :description,
         :publisher_title,
-        :phase,
         :related_discussion_href,
         :related_discussion_title,
         :update_type,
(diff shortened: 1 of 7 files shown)


KushalP:
Is there any reason why we're not testing that the default is set when creating editions?

tadast:
@KushalP I've added a unit test and rebased

PR #47, opened 2015-11-11
Create new edition drafts lazily

tadast:
After trying out creating new draft editions as soon as the user clicks on the
"Create new edition" links  we decided not to use this approach anymore because
views needed to be aware of the current state of the latest edition and show
either a link or a form button. This makes view code more complicated and is
hard to style consistently.

Moving the logic deeper into the back-end makes the code simpler and more
robust (e.g. validations at the model level).

Gotcha: this creates a chance for unlikely race-conditions i.e. if two content
designers click "Create new edition" links at the same time, they will each
create a new draft edition on clicking "Save Draft". This should not be a big
problem as both editions will be visible in history, but only the latest
edition will be editable. Also, publishing-api should soon implement optimistic
locking that would eliminate this problem.

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -41,16 +41,15 @@ class GuidesController < ApplicationController
     @guide = Guide.find(params[:id])
     @comments = @guide.comments_for_rendering
 
-    if @guide.latest_edition.published?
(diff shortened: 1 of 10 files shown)


AndrewVos:
lg

PR #48, opened 2015-11-11
Rename publisher_* to content_owner_*

tadast:
We have realised that using publisher to indicate the group responsible for a
piece of content can be ambiguous because within GOV.UK we use the term
"publisher" to name applications that publish content.

Other repos that need this renaming: govuk-content-schemas, government-frontend

We will not need to sync deploys or update because none of those are used in
production.

The build will go green once https://github.com/alphagov/govuk-content-schemas/pull/150 gets merged

/cc @KushalP

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -1,26 +1,21 @@
 class Edition < ActiveRecord::Base
   acts_as_commentable
 
-  PUBLISHERS = {
-    "Design Community" => "http://sm-11.herokuapp.com/designing-services/design-community/",
-    "Agile Community" => "http://sm-11.herokuapp.com/agile-delivery/agile-community"
-  }.freeze
-
   belongs_to :guide, touch: true
   belongs_to :user
 
   has_many :approvals
 
+  belongs_to :content_owner
+
   scope :draft, -> { where(state: 'draft') }
   scope :published, -> { where(state: 'published') }
   scope :review_requested, -> { where(state: 'review_requested') }
 
-  validates_presence_of [:state, :phase, :description, :title, :update_type, :body, :publisher_title, :publisher_href, :user]
+  validates_presence_of [:state, :phase, :description, :title, :update_type, :body, :content_owner, :user]
   validates_inclusion_of :state, in: %w(draft published review_requested approved)
   validates :change_note, presence: true, if: :major?
   validate :published_cant_change
-
-  before_validation :assign_publisher_href
 
   %w{minor major}.each do |s|
     define_method "#{s}?" do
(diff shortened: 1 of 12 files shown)


app/models/edition.rb
```diff
@@ -1,7 +1,7 @@
 class Edition < ActiveRecord::Base
   acts_as_commentable
 
-  PUBLISHERS = {
+  CONTENT_OWNERS = {
```
AndrewVos: Why not put this in the database?

AndrewVos:
Ok I've merged the content-schemas PR, so this should pass now.

app/models/edition.rb
KushalP: Putting this in the database would be nice for future work.

KushalP:
I kicked off another build of this PR and it looks green. There are some merge conflicts that need to be dealt with before it can be merged in.

app/models/edition.rb
```diff
@@ -1,26 +1,23 @@
...
   scope :review_requested, -> { where(state: 'review_requested') }
 
-  validates_presence_of [:state, :phase, :description, :title, :update_type, :body, :publisher_title, :publisher_href, :user]
+  validates_presence_of [:state, :phase, :description, :title, :update_type, :body, :content_owner, :user]
   validates_inclusion_of :state, in: %w(draft published review_requested approved)
   validates :change_note, presence: true, if: :major?
   validate :published_cant_change
 
-  before_validation :assign_publisher_href
+  accepts_nested_attributes_for :content_owner
```
tadast: I don't think we need this as we do not intend to modify ContentOwner records in the UI

db/seeds.rb
```diff
@@ -46,14 +46,14 @@
...
       edition = Edition.new(
-        title:           title,
-        state:           state.present? ? state : "draft",
-        phase:           "alpha",
-        description:     "Description",
-        update_type:     "minor",
-        body:            body,
-        publisher_title: Edition::PUBLISHERS.keys.first,
-        user:            author
+        title:               title,
```
tadast: Minor: spacing looks a bit funny

spec/support/generators.rb
```diff
@@ -1,15 +1,17 @@
...
-      title:           "The Title",
-      state:           "draft",
-      phase:           "beta",
-      description:     "Description",
-      update_type:     "major",
-      change_note:     "change note",
-      body:            "# Heading",
-      publisher_title: Edition::PUBLISHERS.keys.first,
-      user:            User.new(name: "Generated User")
+      title:               "The Title",
```
tadast: Minor: spacing looks a bit funny

tadast:
Andrew has made changes and rebased my commit, his changes LGTM

PR #49, opened 2015-11-11
Redirect to the guide edition show after approving

tadast:
Keep it consistent with other workflow actions

cc @AndrewVos


AndrewVos:
lg

PR #50, opened 2015-11-12
Allow migrating old guide pages to new guide pages

AndrewVos:
Old guide pages can be redirected one by one to new published guide paths.

Use `rake db:seed` to populate the db with slug migrations.


AndrewVos:
@tadast I have reviewed and actioned all your comments.

app/controllers/slug_migrations_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def update
+    @slug_migration = SlugMigration.find(params[:id])
+    @select_options = Guide.with_published_editions.pluck(:slug, :id)
+
+    slug_migration_parameters = params.require(:slug_migration)
+      .permit(:guide_id)
+    slug_migration_parameters[:completed] = true if params[:save_and_migrate].present?
+
```
tadast: For the below block, what about
```ruby
    if @slug_migration.update_attributes(slug_migration_parameters)
      @selected_guide_id = @slug_migration.guide_id
      if @slug_migration.completed?
        SlugMigrationPublisher.new.process(@slug_migration)
        redirect_to slug_migration_path(@slug_migration), notice: "Slug Migration has been completed"
      else
        redirect_to edit_slug_migration_path(@slug_migration), notice: "Slug Migration has been saved"
      end
    else
      render action: :edit
    end
  end
rescue GdsApi::HTTPClientError => e
  flash[:error] = e.error_details["error"]["message"]
  render action: :edit
end
```

app/controllers/slug_migrations_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    @incompleted_count = SlugMigration.where(completed: false).count
+  end
+
+  def edit
+    @slug_migration = SlugMigration.find(params[:id])
+    @select_options = Guide.with_published_editions.pluck(:slug, :id)
+    @selected_guide_id = @slug_migration.guide_id
+  end
+
+  def update
```
tadast: Perhaps we could define `selected_guide_id` and `select_options` as methods and declare them as `helper_method`, so that the views can access them whenever it's needed. Currently instantiating those variables is duplicated (see edit), makes the controller action code a bit noisier and may not always be required (e.g. if we chose to redirect to show/index on success)

app/models/slug_migration.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  validates :slug, uniqueness: true
+  validate :guide_must_be_nil_or_published
+  validate :guide_cant_be_empty_when_migrating
+  validate :is_not_already_completed
+
+  before_validation on: :create do |object|
+    object.content_id = SecureRandom.uuid
+  end
+
```
tadast: Minor: I like putting validation methods under `private` to reduce the public class API as they should never be called from outside the class.

app/models/slug_migration_publisher.rb
```diff
@@ -0,0 +1,22 @@
...
+    data = {
+      content_id: slug_migration.content_id,
+      format: "redirect",
+      publishing_app: "service-manual-publisher",
+      base_path: slug_migration.slug,
+      redirects: [
+        {
+          path: slug_migration.slug,
+          type: "exact",
+          destination: slug_migration.guide.slug,
```
tadast: hmm, not related to this PR, but I'm surprised that destination is not a content-id

app/views/slug_migrations/edit.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    <h2>Slug Migration <small><%= @slug_migration.slug %></small></h2>
+    <div class="form-group">
+      <%= f.label :guide_id, "Guide" %>
+      <%=
+        f.select :guide_id,
+          options_for_select(@select_options, @selected_guide_id),
+          {include_blank: "Please choose a guide to redirect to"},
+          {class: "slug-migration-select-guide input-md-12 form-control"}
+      %>
+      <%= f.error_list :guide %>
```
tadast: Not sure about this one, should it be `guide` or `guide_id`?

AndrewVos: Nope, it should be `guide` because that's what gets validated.

tadast:
I have a few suggestions from the UX perspective: 
1) when you successfully complete a migration it would be nice to be able to click the actual old link and see it being redirected.
2) Why do we need "Save" in addition to "Save and Migrate"?

Other than that there a minor comments left that you may choose to address or not, but overall it LGTM :+1:

KushalP:
> 1) when you successfully complete a migration it would be nice to be able to click the actual old link and see it being redirected.

This sounds like a great idea.

> 2) Why do we need "Save" in addition to "Save and Migrate"?

A good question. This was an assumption based on how the content designers expect to work. They want to be able to edit the slug for a guide right up to the date it's published. The content designers also wanted to enable redirects based on the "sections" or "communities" they're working on in one go rather than as single pages. With this in mind, allowing edits to the redirect associations right up until they're published made sense.

PR #54, opened 2015-11-16
Add .html to old slug urls when seeding slug migrations

AndrewVos:
All the jekyll urls have .html suffix.


KushalP:
They all work without the `.html` suffix as well, see:

- https://www.gov.uk/service-manual/agile/writing-user-stories
- https://www.gov.uk/service-manual/agile/writing-user-stories.html

tadast:
oooh, does this mean SlugMigration needs an array of slugs instead of a single one?

KushalP:
It doesn't have to. In this case it can register the `.html` suffix if it's not present in the slug field.

PR #55, opened 2015-11-16
Use a reproducible tmp path for old repo download

AndrewVos:
Seed a lot faster because we're not cloning down the old repo every
single time.


tadast:
http://ruby-doc.org/stdlib-1.9.3/libdoc/tmpdir/rdoc/Dir.html#method-c-tmpdir would be slightly nicer but this LGTM too :+1:

PR #56, opened 2015-11-16
Insert slug migrations in a migration

AndrewVos:
Because we need this data in production, and because it never changes,
we instead insert slug migrations in a migration.


tadast:
LGTM. The only improvement I could suggest would be hardcoding content ids alongside hardcoded slugs for consistency within a dev machine - publishing API complains if the same base_path gets posted with a different content id.

PR #57, opened 2015-11-16
Allow markdown in change notes, and auto link urls

AndrewVos:
Content editors often paste urls into change notes, so render change
notes as markdown and use auto linking.

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -69,6 +69,15 @@ class Edition < ActiveRecord::Base
     guide.editions.published.where("id < ?", id).order(id: :desc).first
   end
 
+  def change_note_html
+    Redcarpet::Markdown.new(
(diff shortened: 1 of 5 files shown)


app/models/edition.rb
```diff
@@ -61,6 +61,16 @@ def latest_edition?
     self == guide.latest_edition
   end
 
+  def change_note_html
+    Govspeak::Document.new(change_note).to_html
```
tadast: Was this line left by mistake?

tadast:
If links is all we need maybe https://github.com/tenderlove/rails_autolink would be a lighter solution?

AndrewVos:
It does appear to support bullet points too in publisher, so I'm going to leave the markdown support in.

tadast:
LGTM :+1:

PR #58, opened 2015-11-16
Edition diffs

tadast:
Content designers can see changes in the current or any other historic version
of an edition compared to the previously published edition. This is helpful
because they do not need to read both versions and compare them manually to see
new additions, deletions and changes.

It currently has no design and therefore the markup is not great either. I will
wait for design before making assumptions on what it's going to look like.

### Preview

screenshot 2015-11-16 12 27 05


AndrewVos:
In your screenshot, the fields (other than `body`) don't show `-` and `+`. Why?

app/views/editions/index.html.erb
```diff
@@ -16,6 +17,11 @@
           <%= link_to "See edition", edition %>
         </td>
         <td>
+        <% if edition.previously_published_edition %>
```
AndrewVos: Perhaps it's an anti-pattern to just hide this.

What about disabling it, and having some popup text explaining why it's not visible?

AndrewVos:
Actually, on second thought, maybe `-` and `+` is not what our users would like to see. Perhaps remove them entirely and explain somewhere at the top what red and green mean?

I think maybe developers are used to seeing `+` and `-` but probably not content editors.

PR #59, opened 2015-11-16
Show users who has authored and approved changes

tadast:
screen shot 2015-11-16 at 16 42 03

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -4,7 +4,7 @@ class Edition < ActiveRecord::Base
   belongs_to :guide, touch: true
   belongs_to :user
 
-  has_many :approvals
+  has_one :approval
 
   belongs_to :content_owner
 
--- a/app/views/editions/show.html.erb
+++ b/app/views/editions/show.html.erb
@@ -49,6 +49,24 @@
         <%= @edition.updated_at %>
       </div>
     </div>
+
+    <div class="panel panel-default">
+      <div class="panel-heading">
+        Authors
+      </div>
+      <div class="panel-body">
+        <label>Changes by:</label>
+        <%= @edition.user.name %>
+      </div>
+      <% if @edition.approval %>
+        <div class="panel-body">
+          <label>Approved by:</label>
+          <%= @edition.approval.user.name %>
+        </div>
+      <% end %>
+    </div>
+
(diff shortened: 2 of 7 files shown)


spec/features/guide_publishing_flow_spec.rb
```diff
@@ -155,7 +155,7 @@
         edition = Generators.valid_edition(state: "review_requested", title: "Standups")
         guide = Guide.create!(latest_edition: edition, slug: "/service-manual/something")
 
-        reviewer = User.new(name: "Some User")
+        reviewer = User.new(name: "Keanu Reviews")
```
KushalP: :ok_hand:

app/models/edition.rb
```diff
@@ -4,7 +4,7 @@ class Edition < ActiveRecord::Base
   belongs_to :guide, touch: true
   belongs_to :user
 
-  has_many :approvals
+  has_one :approval
```
AndrewVos: How does this work without migrations? Does it take the last approval?

AndrewVos:
I'm not sure I like that `has_one`.

app/views/editions/show.html.erb
```diff
@@ -49,6 +49,24 @@
         <%= @edition.updated_at %>
       </div>
     </div>
+
+    <div class="panel panel-default">
+      <div class="panel-heading">
+        Authors
```
KushalP: If this is a heading, why isn't it using one of the respective `<h*>` elements?

app/models/edition.rb
tadast: We only had one approval per edition already, so it works nicely without migrating any data.

app/views/editions/show.html.erb
tadast: I followed examples on http://getbootstrap.com/components/#panels. These are very likely to change once Mike goes through the UI (soon), so I wouldn't worry about it too much.

tadast:
@AndrewVos apart from disliking `has_one` do you have any concrete concerns about it?

AndrewVos:
Yeah, the overloaded use of `has_one` is less explicit than I would like it to be. Consider adding `approved_by_id => User`

PR #66, opened 2015-11-19
Redirect back to the previous page after posting a comment

tadast:
To make UX less confusing redirect users to the previous page after posting a comment, so they can see their saved comment on the screen.


AndrewVos:
lg

PR #67, opened 2015-11-19
Add search for edition title and body

AndrewVos:
Add search for edition title and body                                           
                                                                                
- Using postgres tsvector search.                                               
- Title takes precedence over body.                                             
- `rake schema:load doesn't bring in triggers from the development              
  database, so we use the `hairtrigger` gem to handle triggers.


AndrewVos:
Broken because of https://github.com/alphagov/service-manual-publisher/pull/68

app/controllers/guides_controller.rb
```diff
@@ -1,9 +1,17 @@
...
   def index
-    @guides = Guide
-                .includes(latest_edition: :user)
-                .order(updated_at: :desc)
-                .page(params[:page])
+    @guides = if params[:q].present?
+                edition_ids = Edition.search(params[:q]).pluck(:id)
+                Guide
+                  .includes(latest_edition: :user)
+                  .where(editions: {id: edition_ids})
```
tadast: Does it preserve the order in which search results were returned?

app/controllers/guides_controller.rb
```diff
@@ -1,9 +1,17 @@
...
+                  .page(params[:page])
+              else
+                Guide
+                  .includes(latest_edition: :user)
+                  .order(updated_at: :desc)
+                  .page(params[:page])
+              end
```
tadast: What about getting rid of some duplicated logic?

```
@guides = Guide.includes(latest_edition: :user).page(params[:page])
if params[:q].present?
  edition_ids = Edition.search(params[:q]).pluck(:id)
  @guide.where!(editions: {id: edition_ids})
else
  @guides = @guides.order(updated_at: :desc)
end

File: app/models/edition.rb [hunk 3]
```diff
@@ -78,6 +78,20 @@ def change_note_html
     ).render(change_note)
   end
 
+  def self.tsearch_query search_terms
```
tadast: missing parenthesis

app/views/layouts/application.html.erb
```diff
@@ -14,6 +14,17 @@
 <% end %>
 
 <% content_for :content do %>
+  <div class="search">
+    <form action="<%=root_path%>", method="get">
```
tadast: Having a more explicit `guides_path` here would be safer, because the root_url can change to point to another controller action.

spec/features/search_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    visit root_path
+    within ".search" do
+      fill_in "Search", with: "testing"
+      click_button "Search"
+    end
+
+    expect(page).to have_text "Unit Testing"
+    expect(page).to_not have_text "Standups"
+  end
```
tadast: One more spec to prove title weighs more than body would be cool

spec/models/edition_spec.rb
```diff
@@ -166,4 +166,32 @@
...
+      titles.each_with_index do |title, index|
+        edition = Generators.valid_edition(state: "review_requested", title: title)
+        Guide.create!(latest_edition: edition, slug: "/service-manual/#{index}")
+      end
+
+      results = Edition.search("testing").map {|e| e.title}
+      expect(results).to eq ["Unit Testing"]
+    end
+
+    it "prioritises title over body" do
```
tadast: This test shows the field weighting, but the end user can't see it because the controller loses the order when loading guides. I do not know how to solve it though.

tadast:
So now we have triggers in the model itself as well as migrations AND schema.rb. Are they all required? Could you please explain how it works in the commit note?

tadast:
If we do indeed need all those triggers, then perhaps we could extract the whole thing to a class/module, or at least move SQL to a constant

AndrewVos:
The sql in the model describes how the migrations will look when generated. Everytime you update the model migrations you regenerate trigger migrations, which drops and recreates previous triggers.

Read how hairtrigger works.

Extracting the sql will hide behaviour.

app/models/guide.rb
```diff
@@ -21,6 +21,14 @@ def self.with_published_editions
       .uniq
   end
 
+  def self.search(search_terms)
+    ids = Edition.search(search_terms).pluck(:guide_id)
+    order = sanitize_sql_array(
+      ["position(id::text in ?)", ids.join(',')]
```
tadast: nice :+1:

tadast:
LGTM

PR #68, opened 2015-11-19
Delete all SlugMigration rows before every test

AndrewVos:
Slug migrations are populated in migrations, so if we recreate and
migrate the database, all tests would fail. This doesn't work very well
in CI.

### DIFF
--- a/db/migrate/20151116113920_insert_slug_migrations.rb
+++ b/db/migrate/20151116113920_insert_slug_migrations.rb
@@ -1,5 +1,7 @@
 class InsertSlugMigrations < ActiveRecord::Migration
   def change
+    return say("Skipping slug creation in test environment") if Rails.env.test?
+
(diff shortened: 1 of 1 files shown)


tadast:
Should we skip the migration in test env instead? This seems like accepting a problem and undoing it later rather than making it right in the first place

AndrewVos:
Not sure you can skip a migration like that?

AndrewVos:
Also, it's not obvious. If we use some code to skip the migration in the migration file, and I see no slug migrations in test, I'll have a hard time finding out why they're not there.

tadast:
`return say("Skipping slug creation in test environment") if Rails.env.test?` within the migration. 

When would you want to rely on data added by migrations in tests?

AndrewVos:
@tadast Oh cool. Ok I like that much more.

PR #69, opened 2015-11-20
Retain changes in the guide form on failure

tadast:
I've noticed that after a validation error or an API error is triggered, the
form loads a wrong edition object and thus loses all the changes made by user.
This can make people really frustrated.

This change makes sure we always use the same latest_edition object in the form.
We tried this attempt before and reverted, because using a has_one association
would replace the published edition (published editions can not be modified).

To avoid this situation the code now ensures there is an editable draft edition
just before updating the existing latest_edition. In such a case it overwrites
the latest_edition id parameter that comes from the form to ensure the newly
created draft edition is updated.

A new side effect of this behaviour can be observed when an update to a
currently published version of guide fails. Then we persist a draft clone of
the latest published edition in the local database. Users should not experience
any difference in the UX compared to the previous behaviour.

### DIFF
--- a/spec/models/guide_spec.rb
+++ b/spec/models/guide_spec.rb
@@ -1,6 +1,32 @@
 require 'rails_helper'
 
 RSpec.describe Guide do
+  describe "#ensure_draft_exists" do
+    let(:edition) { Generators.valid_published_edition }
+    let(:guide) do
+      Guide.create!(slug: "/service-manual/slug", latest_edition: edition)
+    end
+
+    it "does nothing if the latest edition is in a draft state" do
+      edition.update_attribute(:state, "draft")
+
+      guide.ensure_draft_exists
+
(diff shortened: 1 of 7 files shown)


spec/models/guide_spec.rb
```diff
@@ -1,6 +1,32 @@
...
+    it "returns the latest edition if it's not published" do
+      edition.update_attribute(:state, "draft")
+
+      guide.ensure_draft_exists
+
+      expect(guide.latest_edition).to eq edition
+      expect(guide.editions.count).to eq 1
+    end
+
+    it "returns the latest edition if it's not published" do
```
AndrewVos: These two tests have the same title.

AndrewVos:
I like this. lgtm

AndrewVos:
(rename that one test though)

PR #70, opened 2015-11-20
Show an error summary on top of the form

tadast:
Errors that are not associated to any field visible in the form can occur e.g.
published editions can't change, a change not might be missing when it's hidden
by JS etc. In addition to that, the erroneous field might not be visible on the
screen as it's at the bottom of the form.

Show a summary of all errors on top of the form to make it clear to the user
a failure has occurred and changes were not saved.

screenshot 2015-11-20 12 36 46


AndrewVos:
looks good to mergel

PR #73, opened 2015-11-23
Guides can be filtered by state

AndrewVos:
https://trello.com/c/2txchOit/63-make-guides-easier-to-find-from-the-index-page

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -4,12 +4,19 @@ class GuidesController < ApplicationController
                 Guide.search(params[:q])
                   .includes(latest_edition: :user)
                   .page(params[:page])
+              elsif params[:state].present?
+                @guides = Guide
+                            .with_state(params[:state])
+                            .page(params[:page])
               else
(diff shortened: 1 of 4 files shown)


app/models/guide.rb
```diff
@@ -29,6 +29,10 @@ def self.search(search_terms)
     where(:id => ids).order(order)
   end
 
+  def self.with_state(state)
+    Guide.includes(:latest_edition).where(editions: {state: state})
```
tadast: Hmm, isn't that going to return Guides that have been published in the past but are now in draft if you search for state: 'published'?

app/controllers/guides_controller.rb
```diff
@@ -4,12 +4,19 @@ def index
                 Guide.search(params[:q])
                   .includes(latest_edition: :user)
                   .page(params[:page])
+              elsif params[:state].present?
```
tadast: Would it be possible to filter out search results?

tadast:
Some tests would be cool

PR #76, opened 2015-11-24
Guides can be filtered by state, user, content owner, or keyword.

AndrewVos:


app/models/guide.rb
```diff
@@ -24,11 +24,15 @@ def self.with_published_editions
   def self.search(search_terms)
     ids = Edition.search(search_terms).pluck(:guide_id)
     order = sanitize_sql_array(
-      ["position(id::text in ?)", ids.join(',')]
+      ["position(guides.id::text in ?)", ids.join(',')]
     )
     where(:id => ids).order(order)
   end
 
+  def self.with_state(state)
```
tadast: This method does not seem to be used anywhere

tadast:
It looks good generally. A few comments:

1) Ideally this would be split into two commits, one for switching to sql schema, one for adding the feature
2) We use "Published by" and "Content Owner" as synonyms. I think it would be nice to keep calling it :content_owners at the code level and only use the term "Published by" on for labels.
3) Would be cool to have a test that proves editions other than the latest edition do not get searched e.g. you have a guide that has been published in the past but currently has a draft state. It should not show up when searching for 'published' guides etc.

PR #77, opened 2015-11-24
Edit form UI

tadast:
Suggestion from our designer:

screen_shot_2015-11-23_at_12 14 10

Current state:
screen shot 2015-11-24 at 16 17 25

## Summary

- The form view is now consistent with other views and uses 8-4 column layout
- The side panel has links to content guidance and formatting help
- Related discussion fields have been removed from the form (TODO: remove
  them completely)
- Fields have been re-arranged, labels and help text added.

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -14,7 +14,7 @@ class GuidesController < ApplicationController
 
   def new
     @guide = Guide.new(slug: "/service-manual/")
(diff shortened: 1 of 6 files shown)


tadast:
@gemmaleigh would you be able to help me with radio button styles for update type? I've seen this style in smart-answers, but I'm not sure if we should be using `static` or something else to get the styles?

PR #80, opened 2015-11-25
Normalise newlines in seed.rb

tadast:
When importing data from the old service manual guides the newline symbols are
"\n", but when saved again via the application they change to "\r\n". This
causes "compare changes" view to highlight the whole document first time it's
saved. Importing data with "\r\n" as newlines solves this problem.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -72,7 +72,7 @@ if Rails.env.development? || ENV["GOVUK_APP_DOMAIN"] == "preview.alphagov.co.uk"
       phase:           "beta",
       description:     "Description",
       update_type:     "minor",
-      body:            object[:body],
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -72,7 +72,7 @@ def all_old_guides
       phase:           "beta",
       description:     "Description",
       update_type:     "minor",
-      body:            object[:body],
+      body:            object[:body].gsub(/\n/, "\r\n"),
```
AndrewVos: Shouldn't this be `.gsub /\r\n?/, "\n"`

tadast: No, rails app saves it with \r\n. This fixes the problem, tested locally

PR #82, opened 2015-11-27
Stop using "Create new edition"

AndrewVos:
It's a concept that users don't need to know, and it confuses them.
This comes from user testing.


tadast:
LGTM

PR #83, opened 2015-11-27
Move comments to a Conversation tab

AndrewVos:
User testing found that the Comments section interfered with the UI too
much, and people didn't know what it was for.

Also, when saving a comment does the guide get saved or not? This is
confusing for the user.


tadast:
LGTM

PR #86, opened 2015-11-27
[do-not-merge] Show a progress bar for edition state

AndrewVos:
Awaiting feedback on this. Perhaps some copy would be nice.

In user testing, some users didn't quite realise there was a pipeline for publishing state. This is an attempt to fix this.

### DIFF
--- a/app/helpers/guide_helper.rb
+++ b/app/helpers/guide_helper.rb
@@ -7,6 +7,25 @@ module GuideHelper
     "published"        => "info",
   }
 
+  def state_progress_bar(guide)
+    state = guide.latest_edition.state
+    return "" if state.nil?
+
+    states = %w(draft review_requested approved published)
+
+    html = '<div class="progress">'
+    states.each do |s|
+      html << <<-HTML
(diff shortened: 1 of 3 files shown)


joshski:
I assume the coloured bit at the top is what you mean by "progress bar"? It would look more like a progress bar of there was some sense of direction, like arrows or step numbers.

Also the names of each state should probably be humanised like "Review Requested"

app/helpers/guide_helper.rb
```diff
@@ -7,6 +7,25 @@ module GuideHelper
...
+  def state_progress_bar(guide)
+    state = guide.latest_edition.state
+    return "" if state.nil?
+
+    states = %w(draft review_requested approved published)
+
+    html = '<div class="progress">'
+    states.each do |s|
+      html << <<-HTML
+                  <div class="progress-bar progress-bar-#{STATE_CSS_CLASSES.fetch(s)}" style="width: 25%">
```
Jetski5822: Style tags? Hello 2001

AndrewVos:
Thanks @joshski nice catch

PR #89, opened 2015-11-30
[do-not-merge] Add guide actions and explanation of state to top of page

AndrewVos:
This shows the user what state the guide is in, and allows them to easily
review/publish it.

image
image
image
image


AndrewVos:
This is not going to be used anymore: https://trello.com/c/QAu4Eggt/96-fix-create-a-new-guide-edition-workflow

PR #90, opened 2015-11-30
Redirect back to guide index after state change

AndrewVos:
When a user changes state to approved/review_requests they probably
wouldn't edit the edition anymore, because they are awaiting a response
from a different actor.

Also, fix Action links in the guide index page.. Publish and Review were
pointing at the edit path, when they should have been pointing at the
show path.

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -1,11 +1,5 @@
 class GuidesController < ApplicationController
   def index
-    @state_titles = {
-      draft: "Edit",
-      review_requested: "Review guide",
-      approved: "Publish",
-      published: "Edit",
-    }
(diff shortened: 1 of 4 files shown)


app/controllers/guides_controller.rb
```diff
@@ -1,11 +1,5 @@
 class GuidesController < ApplicationController
   def index
-    @state_titles = {
-      draft: "Edit",
-      review_requested: "Review guide",
-      approved: "Publish",
-      published: "Edit",
-    }
     @user_options = User.all.collect{ |u| [u.name, u.id] }
```
mobaig: unrelated, but use `pluck` here instead?

mobaig:
apart from failing tests, this LGTM

PR #92, opened 2015-12-01
Index guides in rummager

AndrewVos:
We will use the `design-principles` frontend to render search results for
now. `design-principles` uses a filter for search results called
`filter_manual`, which works with the old guide service manual.

image
image
image

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -97,6 +97,7 @@ private
     ActiveRecord::Base.transaction do
       @guide.latest_edition.update_attributes!(state: 'published')
       GuidePublisher.new(guide: @guide).publish
+      SearchIndexer.new(@guide).index
       redirect_to back_or_default, notice: "Guide has been published"
     end
   rescue GdsApi::HTTPErrorResponse => e
--- a/lib/tasks/rummager.rake
+++ b/lib/tasks/rummager.rake
@@ -1,0 +1,39 @@
+namespace :rummager do
+  desc "index all published documents in rummager"
+  task index: :environment do
+    require 'plek'
+    require 'rummageable'
+
+    root_document = {
+      format: "manual",
+      title: "Government Service Design Manual",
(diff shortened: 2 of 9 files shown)


lib/tasks/rummager.rake
```diff
@@ -0,0 +1,31 @@
...
+      edition = guide.latest_edition
+      if edition.published?
+        puts "Indexing #{edition.title}..."
+        {
+          "_type": "manual_section",
+          "description": edition.description,
+          "indexable_content": edition.body,
+          "title": edition.title,
+          "link": guide.slug,
+          "manual": "service-manual",
```
rboulton: This field should have a leading slash (ie `/service-manual`).  It is expected to contain the base path of the manual.  Eg, existing values we have are "/guidance/digital-and-technology-skills".

rboulton:
This looks pretty good.

You should also send a single document from the task, representing the manual as a whole.  This looks like:

```
{
  format: "manual",
  title: "Government Service Design Manual",
  description: "All new digital services from the government must meet the Digital by Default Service Standard",
  link: "/service-manual",
  organisations: "government-digital-service",
}
```

This will keep the data in rummager consistent, and allow the standard filtered manual search results page to work (ie, https://www.gov.uk/service-manual).

AndrewVos:
@rboulton Thanks, I took your changes on board.

Is "_type" required?

app/controllers/guides_controller.rb
```diff
@@ -97,6 +97,7 @@ def publish
     ActiveRecord::Base.transaction do
       @guide.latest_edition.update_attributes!(state: 'published')
       GuidePublisher.new(guide: @guide).publish
+      SearchIndexer.new(@guide).index
```
tadast: Do we need to think about the exceptions this may raise, as it will roll back the transaction?

AndrewVos: Yeah, I've created a trello ticket for this.

PR #95, opened 2015-12-03
Show a disabled "send for review" button when not persisted

AndrewVos:
The "Send for review" button is now displayed when creating a new guide,
because
https://trello.com/c/QAu4Eggt/96-fix-create-a-new-guide-edition-workflow

Also, hide other guide action buttons instead of disabling them.


mobaig:
:shipit:

PR #96, opened 2015-12-03
Change titles of edition action buttons

AndrewVos:
- Don't add "Back to all guides" button, because this essentially
replicates the behaviour of the browser back button.
- Don't have different save buttons for persisted and not persisted
  guides, because it's inconsistent.

https://trello.com/c/YoTKvG78/87-renaming-and-restructuring-calls-to-action


mobaig:
:rocket:

PR #101, opened 2015-12-11
Action buttons. Action buttons everywhere

tadast:
Duplicate all guide action buttons at the bottom of the page.

As per design suggestions - this should make it easier for the users to find
action buttons regardless of where they are on the page.

Many tests broke because button matchers became ambiguous with two of each
button visible on the page. I've written a helper that allows to click the
first visible button.


AndrewVos:
Looks good

PR #102, opened 2015-12-14
Revamp change notes UI

tadast:
After user testing it has became clear that users do not understand the two
modes of update types. Make it clear what each mode means in the UI.

Also, move the whole thing to the top of the form and visually separate it from
the content section (I've put in in a separate 'well' container). This implies
that users should write the change note before making any content editions.

Add a 'change summary' field alongside the 'change note' field. This should
make users think about how they write change notes and suggest git/github
change note style.


AndrewVos:
lgtm

PR #103, opened 2015-12-14
Fix a test that started failing due to a merge conflict

tadast:
The issue was introduced by combining changes from the search branch and the
branch that added extra row of action buttons.

@AndrewVos


AndrewVos:
lgtm

PR #104, opened 2015-12-14
Disable publishing based on an ENV variable

tadast:
We intend to deploy the application to production but we are not ready to
publish any documents yet (we need to coordinate old content replacement,
prepare topic pages etc.). Having the application in production will allow
content designers to work on drafts without losing them every night (as in
preview/integration environment).

We intend to set this environment variable in production instances only (not
staging).

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -94,6 +94,12 @@ private
   end
 
   def publish
+    if ENV['DISABLE_PUBLISHING']
+      @edition = @guide.latest_edition
+      flash[:error] = "Publishing is currently disabled. The guide has not been published."
+      return render template: 'editions/show'
+    end
+
     ActiveRecord::Base.transaction do
       @guide.latest_edition.update_attributes!(state: 'published')
       GuidePublisher.new(guide: @guide).publish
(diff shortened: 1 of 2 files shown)


tadast:
Build failure should be resolved once #103 gets merged

AndrewVos:
Looks good to me. Though, I feel like it should be `ENABLE_PUBLISHING` instead.

app/controllers/guides_controller.rb
```diff
@@ -94,6 +94,12 @@ def approve_for_publication
   end
 
   def publish
+    if ENV['DISABLE_PUBLISHING']
+      @edition = @guide.latest_edition
+      flash[:error] = "Publishing is currently disabled. The guide has not been published."
+      render template: 'editions/show' and return
```
AndrewVos: `return render template: 'editions/show'` ?

tadast:
I was considering ENABLE_PUBLISHING too, but that would mean we'd need to add it to every environment in puppet config (test, dev, preview, staging). I feel like normal behaviour is having this enabled and it's cleaner to just have this variable in a single environment.

tadast:
@AndrewVos I've addressed your comments :)

AndrewVos:
:ship:

PR #107, opened 2015-12-16
Add non html suffix paths as slug migrations too

AndrewVos:
The old service manual sees /service-manual/agile.html and
/service-manual/agile as the same thing.


tadast:
Does this mean people will have to be aware of the fact that they have to migrate both `.html` and non-`.html` slugs? Can't we just automate it by making two requests to publishing-api each time we migrate?

AndrewVos:
Yeah, you're probably right.

AndrewVos:
Although that approach means having an alternate content_id and url path for each item. It's a much larger, more hacky change.

Let's leave it like this.

tadast:
Based on a slack convo this LGTM :+1:

txnmsgyx-2015 12 16-13-53-15

PR #108, opened 2015-12-16
Improve diffing UI based on user feedback

tadast:
The results from the content design survey weren't totally conclusive but the
eventual winner was "+/- at the start of the rows with a strike through for
deletions".

The only point of confusion around this option was the +/- sit a little too
close to the rest of the body copy, and thus looks a bit like the markdown
itself. Increase the margin between the two.

lwklgajw-2015 12 16-16-05-57

### DIFF
--- a/app/assets/stylesheets/diffy.scss.erb
+++ b/app/assets/stylesheets/diffy.scss.erb
@@ -1,7 +1,19 @@
 <%= Diffy::CSS %>
 
 .diff {
+  $symbol-spacing: 1.5em;
+
+  del {
+    text-decoration: line-through;
+  }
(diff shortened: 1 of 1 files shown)


AndrewVos:
lgtm

AndrewVos:
Hey have you seen this? https://github.com/samg/diffy/blob/master/lib/diffy/css.rb#L18

tadast:
It looks like this:
wlpsudtw-2015 12 16-16-53-31

Given taht this option was not included in the survey and people are currently using GitHub I'd say let's park it for now and revisit if there are any issues.

PR #109, opened 2015-12-17
Use correct "manual" attribute for guide search index data

AndrewVos:


rboulton:
I don't think this is correct - all our manual fields elsewhere have a leading '/'.  For example: https://www.gov.uk/api/search.json?filter_link=/guidance/digital-and-technology-skills/design&fields=format,link,organisations,title,manual

What was the brokenness that this change was made to fix?

AndrewVos:
@rboulton Search wasn't actually working for us. I looked in `design-principles` and things are being indexed using "service-manual" without the leading slash.

AndrewVos:
@rboulton  https://github.com/alphagov/design-principles/blob/master/app/controllers/search_controller.rb#L18

PR #115, opened 2015-12-22
Users can delete search documents from search index for slug migrations

AndrewVos:
This allows users to delete search indexes for the old service manual,
at their leisure.

~~We should update `gds-api-adapters` to the latest version when this gets a gem release: https://github.com/alphagov/gds-api-adapters/pull/413~~

image

image

### DIFF
--- a/app/controllers/slug_migrations_controller.rb
+++ b/app/controllers/slug_migrations_controller.rb
@@ -46,4 +46,11 @@ class SlugMigrationsController < ApplicationController
   def show
     @slug_migration = SlugMigration.find(params[:id])
   end
+
+  def delete_search_index
+    @slug_migration = SlugMigration.find(params[:slug_migration_id])
+    @slug_migration.delete_search_document!
+
+    redirect_to edit_slug_migration_path(@slug_migration), notice: "Document has been removed from search"
+  end
 end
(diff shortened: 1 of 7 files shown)


rboulton:
I think the description of this PR should be to allow deleting documents from the search index, not search indexes themselves?  I'd note that the search-admin tool already has an interface for doing this, but having a more convenient one here is fine.

What prevents the document just coming back the next time the old service manual reindex jenkins job is run?  It gets run automatically from time to time, if I remember rightly.

tadast:
Hi @rboulton, it's part 1 of two of this effort. I've written this up on trello here https://trello.com/c/13B2AoSc/275-come-up-with-a-plan-to-migrate-search-indexes

tadast:
Code looks good to me

app/controllers/slug_migrations_controller.rb
```diff
@@ -46,4 +46,11 @@ def update
   def show
     @slug_migration = SlugMigration.find(params[:id])
   end
+
+  def delete_search_index
+    @slug_migration = SlugMigration.find(params[:slug_migration_id])
+    @slug_migration.delete_search_document!
+
+    redirect_to edit_slug_migration_path(@slug_migration), notice: "Search index has been deleted"
```
tadast: As Richard has mentioned, we should say: "Document has been deleted from the search index" or something similar

PR #116, opened 2015-12-23
Disable publish/approve/review actions when form is modified

AndrewVos:
image

I'm not entirely sure I like this approach. Could the reviewer focus on UX too.

Just noticed some issues:
- doesn't work with radio buttons
- event gets attached to form multiple times

https://trello.com/c/Gflw1nRP/296-disable-the-send-for-review-button-when-the-edit-form-is-dirty

### DIFF
--- a/app/assets/javascripts/disable-when-dirty.js
+++ b/app/assets/javascripts/disable-when-dirty.js
@@ -1,0 +1,12 @@
+$(function() {
+  $("form").each(function() {
+    var form = $(this);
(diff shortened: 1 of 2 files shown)


tadast:
I've implemented an alternative solution to this problem in https://github.com/alphagov/service-manual-publisher/pull/117 that addresses some of the issues this approach has. What do you think?

AndrewVos:
Sounds good to me. Closing this.

PR #117, opened 2015-12-24
Notify users when there are unsaved changes

tadast:
Ask users to confirm their navigation away when the guides form has unsaved
changes. It works by recording a checksum of the form values on page load, then
checking if the checksum has changed before the user tries to navigate away.
This is better than relying on 'change' events of form inputs because it takes
into account the fact that users may change the values back to the previous
ones.

~~A known shortcoming: elements that navigate away have to be explicitly marked
with a css class '.js-disable-when-dirty'. If a user navigates away clicking on
other links this will not trigger. I've tried using 'onbeforeunload' event, but
it also triggers when "Save" and "Save and preview" buttons are clicked - I am
not sure how to make it work without using ugly hacks.~~

closes #116 

If we're happy with this approach, should we add some (PhantomJS?) tests?

shwnpwmi-2016 01 04-11-29-25

### DIFF
--- a/app/assets/javascripts/protect_form_data.js
+++ b/app/assets/javascripts/protect_form_data.js
@@ -1,0 +1,50 @@
+$(function(){
+  window.okToNavigateAway = false;
+
+  $("form.js-protect-data").each(function(_, form) {
+    var $form = $(form);
+    recordChecksum($form);
+    $form.on("click", ".js-ok-to-navigate-away", function(evt){
+      window.okToNavigateAway = true;
+    });
+  });
+
+  window.onbeforeunload = function() {
+    if (window.okToNavigateAway){
+      return undefined;
+    }
+
+    forms = $("form.js-protect-data");
+    for (var i = forms.length - 1; i >= 0; i--) {
+      var form = forms[i];
+      if (hasChanged($(form))) {
+        return "It looks like there are some unsaved changes in the form.\n\nLeaving the page will discard all unsaved changes.";
+      }
+    }
+  };
+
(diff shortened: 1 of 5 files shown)


AndrewVos:
Maybe don't use the word "dirty", it's a bit weird. I like changed more.

AndrewVos:
I'm not sure how I feel about popups saying you've changed the form.

Maybe we should speak to a UX person and find out what's in vogue.

Perhaps autosaving documents might be a better choice.

AndrewVos:
https://www.nngroup.com/articles/ten-usability-heuristics/

AndrewVos:
How about http://stackoverflow.com/a/11766745/128167

tadast:
I've switched to using `onbeforeunload` and whitelisting "Save" and "Save and Preview" buttons instead of explicitly marking all ways to navigate away. This works with reloading a page too now.



Can you have another look please @AndrewVos?

app/assets/javascripts/protect_form_data.js
```diff
@@ -0,0 +1,50 @@
...
+  if (string.length === 0) return hashValue;
+  for (i = 0, len = string.length; i < len; i++) {
+    chr   = string.charCodeAt(i);
+    hashValue  = ((hashValue << 5) - hashValue) + chr;
+    hashValue |= 0; // Convert to 32bit integer
+  }
+  return hashValue;
+}
+
+function recordChecksum(form) {
```
AndrewVos: This is pretty simply stuff, maybe it doesn't need a global function?

PR #118, opened 2015-12-24
[Discuss] Opening preview links in a new tab

tadast:
User research and early user feedback has shown that vast majority of the users
expect this to open in a new tab. Some users found it difficult to navigate
back to the form when we open the preview window in the same tab.

The "Preview" link is simple, adding `target="_blank"` solves it.

However, "Save and Preview" is a bit more challenging. To save the draft first we need to submit the form. Unfortunately buttons do not have a `target="_blank"` equivalent. After trying a few hackish
solutions I've come to the conclusion that it is best to:

1) Submit the form via Ajax
2.1) If the save succeeds, proceed opening the preview link in a new tab
2.2) If the save fails, prevent the new tab opening and submit the form once
     again, this time without Ajax so that the user can see a validation message

There is a slight downside when the save fails and the user fixes the errors
and tries to "Save and Preview" once more: since the saving happens via Ajax,
validation messages do not disappear. Any ideas how to prevent this?

If we think this approach is the best there is, we should add some tests and make sure non-javascript version uses the previous behaviour (progressive enhancement etc etc).

### DIFF
--- a/app/assets/javascripts/save_before_continuing.js
+++ b/app/assets/javascripts/save_before_continuing.js
@@ -1,0 +1,23 @@
+$(function() {
+  $(document).on("click", ".js-save-before-continuing", function(event){
+    var link = $(event.currentTarget);
+    var form = link.closest('form');
+    var serialisedForm = form.serialize();
+
+    $.ajax({
+      async: false, // wait for this to finish proceeding with the event
+      type: form.attr('method').toUpperCase(),
+      url: form.attr('action'),
+      data: form.serialize(),
+      complete: function(response) {
+        if (response.status == 200){
+          return true;
+        } else {
+          event.preventDefault();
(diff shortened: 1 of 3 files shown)


tadast:
From an offline discussion:

Andrew thinks this is quite brittle and suggests we have "Preview" link only that gets disabled as soon as there are unsaved changes in the form. I agree JS would be simpler, but it would degrade user experience a bit. I'll talk to a few users about this.

tadast:
@AndrewVos actually with https://github.com/alphagov/service-manual-publisher/pull/117 in place users will get notified if they are trying to preview unsaved document. I'll open a new PR

tadast:
`onbeforeunload` does not trigger when opening a link in a new tab so #117 is not going to help :(

PR #119, opened 2015-12-28
Email notifications for important events

tadast:
Send an email notification to the edition author when:

- Somebody comments on the edition
- Somebody approves the draft for publishing
- Somebody publishes the draft

I'd like the reviewer to pay attention to the copy in emails as well as the code changes :love_letter: 

Should we add Sidekiq or something similar to process the emails? Async job queue would also be useful for search indexing.

Before merging:

- [x] Ensure we have Amazon SES configured for delivering emails in preview, staging and production
- [x] Add a queue processor or decide not to for short term


tadast:
@bradwright has opened a PR to symlink mailer config in deployment scripts https://github.gds/gds/alphagov-deployment/pull/1130

@AndrewVos could you please have a look at the code?

app/controllers/comments_controller.rb
```diff
@@ -6,6 +6,10 @@ def create
       comment: params[:comment][:comment],
     )
 
+    unless edition.notification_subscribers == [comment.user]
+      NotificationMailer.comment_added(comment).deliver_later
```
AndrewVos: What's deliver_later? Apparently it uses "Active Job". Is this new?

AndrewVos: oh:

```
4 Job Execution
For enqueuing and executing jobs you need to set up a queuing backend, that is to say you need to decide for a 3rd-party queuing library that Rails should use. Rails itself does not provide a sophisticated queuing system and just executes the job immediately if no adapter is set.


```

tadast: It's a really cool new Rails API, defaults to synchronous job execution but as soon as we add a queue processor it will do it async without any change to the code!

AndrewVos: Thereby breaking all our tests :)

tadast: True that. Still better than older Rails :)

PR #122, opened 2016-01-05
Automatically generate url slug from guide title

AndrewVos:
Still allow editor to change slug while creating guides though.

### DIFF
--- a/spec/features/guide_spec.rb
+++ b/spec/features/guide_spec.rb
@@ -236,6 +236,30 @@ RSpec.describe "creating guides", type: :feature do
     end
   end
 
+  describe "slug generation" do
+    it "generates slug", js: true do
+      {
+        "Guide title": "guide-title",
+        "slug--with-----hyphens": "slug-with-hyphens",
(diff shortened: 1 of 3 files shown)


spec/features/guide_spec.rb
```diff
@@ -236,6 +236,30 @@
     end
   end
 
+  describe "slug generation" do
```
tadast: perhaps add `js: true` here, because you've forgotten it for the test on lines 254-260 :)

tadast:
Other than my comment above it LGTM

spec/features/guide_spec.rb
```diff
@@ -236,6 +236,30 @@
...
+    it "generates slug", js: true do
+      {
+        "Guide title": "guide-title",
+        "slug--with-----hyphens": "slug-with-hyphens",
+        "       space    slugs  ": "space-slugs",
```
tadast: whatcha gonna do
whatcha gonna do when we come for you

PR #123, opened 2016-01-05
Stricter validation for slugs

tadast:
~~I've noticed that one of the draft guides was added with a slug containing uppercase letters. We should avoid that. Also~~ add validation to prevent weird characters in slugs.

### DIFF
--- a/app/models/guide.rb
+++ b/app/models/guide.rb
@@ -61,6 +61,8 @@ private
       errors.add(:slug, "must be present and start with '/service-manual/'")
     elsif !slug.to_s.match(/\A\/service-manual\/\w+/)
       errors.add(:slug, "must be filled in")
+    elsif !slug.to_s.match(/\A\/service-manual\/[a-z0-9\-\/]+$/i)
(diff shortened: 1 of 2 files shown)


app/models/guide.rb
```diff
@@ -12,6 +12,7 @@ class Guide < ActiveRecord::Base
 
   before_validation on: :create do |object|
     object.content_id = SecureRandom.uuid
+    slug.try(:downcase!)
```
AndrewVos: Slug should never be nil. Effectively this is like doing `slug.downcase! rescue nil`.

tadast:
From https://insidegovuk.blog.gov.uk/url-standards-for-gov-uk/
> GOV.UK URLs are lower case by default. However if there is strong evidence of a user need for  an upper-case short URL (for example, it will be read aloud and refers to a commonly-capitalised phrase, like www.gov.uk/P800) this can be set up separately.

I guess we should not lowercase slugs then...?

tadast:
I've removed the automated lowercasing. https://github.com/alphagov/service-manual-publisher/pull/122 should help with that. We can still merge the stricter validation though

PR #129, opened 2016-01-08
Image uploading and embedding

tadast:
GitHub-style image uploading, yayy. Uses github.com/alphagov/asset-manager API to handle file uploads.

## Gif demo

upload

## Concerns

Is checking file mime type enough to avoid security issues and prevent errors that can be prevented?

Should we build a fallback for users with touch-screen devices?

## Todo

- [x] Finish and merge https://github.com/alphagov/govuk-puppet/pull/3900

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,3 +4,4 @@
 require File.expand_path('../config/application', __FILE__)
 
 Rails.application.load_tasks
+task default: "jasmine:ci"
--- a/app/assets/javascripts/image-upload-markdown.js
+++ b/app/assets/javascripts/image-upload-markdown.js
@@ -1,0 +1,74 @@
+// This is a modified version of https://github.com/stuartnelson3/blog/blob/master/public/js/script.js
+
+function ImageUploadMarkdown(event){
+  var that = this;
+
+  this.init = function(){
+    event.preventDefault();
+    this.$textarea = $(event.currentTarget);
+    var dataTransfer = event.originalEvent.dataTransfer;
+
+    this.addUploadPlaceholders(dataTransfer);
+    for (var i = 0; i < dataTransfer.files.length; i++) {
+      var file = dataTransfer.files[i];
+      this.handleFileUpload(file);
+    }
+  };
+
+  this.addUploadPlaceholders = function(dataTransfer) {
+    var placeholders = [];
+    for (var i = 0; i < dataTransfer.files.length; i++) {
+      var fileName = dataTransfer.files[i].name;
+      placeholders.push(this.placeholderValue(fileName));
+    }
+
+    var selectionStart = this.$textarea[0].selectionStart;
+    var uploadsPlaceholder = placeholders.join("\r\n");
+    this.$textarea.val(
+      [this.$textarea.val().slice(0, selectionStart), uploadsPlaceholder, this.$textarea.val().slice(selectionStart)].join('')
+    );
+  };
+
+  this.replaceImageMarkdownPlaceholder = function(newValue, fileName){
+    var oldPlaceholder = this.placeholderValue(fileName);
+    this.$textarea.val(
+      this.$textarea.val().replace(oldPlaceholder, newValue)
+    );
+  };
+
+  this.placeholderValue = function(fileName) {
+    return "![Uploading " + fileName + "...]()";
+  };
+
+  this.handleFileUpload = function(file) {
+    var formData = new FormData();
+    formData.append("format", "js");
+    formData.append("file", file);
+
+    formData.append("authenticity_token", $('meta[name="csrf-token"]').attr("content"));
+
+    var request = new XMLHttpRequest();
+    request.open("POST", "/uploads");
+
+    request.onloadend = function(fileName) {
+      return function(e) {
+        if(e.currentTarget.status == 201){
+          var filePath = e.currentTarget.response;
+          that.replaceImageMarkdownPlaceholder("!["+fileName+"]("+filePath+")", fileName);
+        } else {
+          that.replaceImageMarkdownPlaceholder("", fileName);
+          window.alert("File upload failure\n\n" + e.currentTarget.response);
+        }
+        that.$textarea.keyup();
+      };
+    }(file.name);
+
+    request.send(formData);
+  };
+
+  return this;
+}
+
+$(document).on('drop', '.js-markdown-image-upload', function(e) {
+  ImageUploadMarkdown(e).init();
+});
--- a/spec/javascripts/image-upload-markdown-spec.js
+++ b/spec/javascripts/image-upload-markdown-spec.js
@@ -1,0 +1,64 @@
+describe("Drag and drop image upload", function() {
+  "use strict";
+
+  var uploadableTextarea,
+      mockDropEvent;
+
+  beforeEach(function() {
+    uploadableTextarea = $("<textarea class='js-markdown-image-upload'>First paragraph\n\nSecond paragraph</textarea>");
+
+    $('body').append(uploadableTextarea);
+    setCursorAfter("First paragraph\n");
+
+    mockDropEvent = {
+      preventDefault: function(){},
+      currentTarget: $('body .js-markdown-image-upload'),
+      originalEvent: {
+        dataTransfer: {
+          files: [{ name: "nice-pic.jpg"}]
+        }
+      }
+    };
+
+    jasmine.Ajax.install();
+  });
+
+  afterEach(function() {
+    jasmine.Ajax.uninstall();
+    uploadableTextarea.remove();
+  });
+
+  it("insert markdown of an uploaded image at the current cursor position", function() {
+    jasmine.Ajax.stubRequest('/uploads').andReturn({
+      response: 'http://asset-api.dev.gov.uk/nice-pic.jpg',
+      status: '201'
+    });
+
+    var upload = new ImageUploadMarkdown(mockDropEvent);
+    upload.init();
+    var markdown = "First paragraph\n![nice-pic.jpg](http://asset-api.dev.gov.uk/nice-pic.jpg)\nSecond paragraph";
+    expect($('body .js-markdown-image-upload').val()).toEqual(markdown);
+  });
+
+  it("show an alert and do not insert any markdown if upload fails", function() {
+    jasmine.Ajax.stubRequest('/uploads').andReturn({
+      response: 'File is too big',
+      status: '422'
+    });
+    spyOn(window, 'alert');
+
+    var upload = new ImageUploadMarkdown(mockDropEvent);
+    upload.init();
+    var markdown = "First paragraph\n\nSecond paragraph";
+
+    expect(window.alert).toHaveBeenCalled();
+    expect($('body .js-markdown-image-upload').val()).toEqual(markdown);
+  });
+
+  function setCursorAfter(value){
+    var textarea = $('body textarea')[0];
+    var cursorAt = value.length;
+    textarea.selectionStart = cursorAt;
+    textarea.selectionEnd = cursorAt;
+  }
+});
--- a/Gemfile
+++ b/Gemfile
@@ -37,6 +37,7 @@ group :development, :test do
   gem 'pry'
   gem 'pry-nav'
   gem 'govuk-content-schema-test-helpers', '~> 1.3.0'
+  gem 'jasmine'
 end
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -115,6 +115,12 @@ GEM
     http-cookie (1.0.2)
       domain_name (~> 0.5)
     i18n (0.7.0)
+    jasmine (2.4.0)
+      jasmine-core (~> 2.4)
+      phantomjs
+      rack (>= 1.2.1)
+      rake
+    jasmine-core (2.4.1)
     jquery-rails (3.1.4)
       railties (>= 3.0, < 5.0)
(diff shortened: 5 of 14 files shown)


app/assets/javascripts/image-upload-markdown.js
```diff
@@ -0,0 +1,60 @@
...
+  e.preventDefault();
+  var dt = e.originalEvent.dataTransfer;
+  for (var i = 0; i < dt.files.length; i++) {
+    var file = dt.files[i];
+    var textarea = e.currentTarget;
+    handleFileUpload(textarea, file);
+  }
+
+  function addImageMarkdownPlaceholder(e, filename) {
+    var startPos = e.selectionStart;
```
AndrewVos: This is not supported in IE8. Do we care?

app/assets/javascripts/image-upload-markdown.js
```diff
@@ -0,0 +1,60 @@
...
+    var fd = new FormData();
+    fd.append("format", "js");
+    fd.append("file", file);
+
+    // CSRF token
+    fd.append("authenticity_token", $('meta[name="csrf-token"]').attr("content"));
+
+    var request = new XMLHttpRequest();
+    request.open("POST", "/uploads");
+
```
AndrewVos: Would be nice to `disable` the textarea until `onloadend`, so the user can't inadvertently break the upload code.

app/assets/javascripts/image-upload-markdown.js
AndrewVos: Sorry, disregard that, I just realised everything here is HTML5 :)

app/assets/javascripts/image-upload-markdown.js
```diff
@@ -0,0 +1,60 @@
...
+    request.onloadend = function(filename) {
```
AndrewVos: `request.addEventListener("progress", updateProgress);` :grin:

app/assets/javascripts/image-upload-markdown.js
tadast: Not sure if I like this idea. This error is easy to recover from (just re-upload again), disabling the text area can be rather annoying.

Also, it would be rather difficult to do as this supports multiple file upload. Maybe that's the main reason why I'm so reluctant :|

AndrewVos: Fair enough, but what if the user changes the upload text while it's happening... Another way might be to just upload the file in the background and immediately set the image to the one that it's going to be?

app/assets/javascripts/image-upload-markdown.js
```diff
@@ -0,0 +1,60 @@
...
+        if(e.currentTarget.status == 201){
+          var filePath = e.currentTarget.response;
+          // Update the text with the correct path for the image.
+          replaceImageMarkdownPlaceholder(textarea, "!["+filename+"]("+filePath+")", filename);
+        } else {
+          // Remove the placeholder markup and show alert
+          replaceImageMarkdownPlaceholder(textarea, "", filename);
+          alert("File upload failure\n\n" + e.currentTarget.response);
+        }
+        $(textarea).keyup();
```
AndrewVos: Why?

app/assets/javascripts/image-upload-markdown.js
```diff
@@ -0,0 +1,60 @@
+// This is a modified version of https://github.com/stuartnelson3/blog/blob/master/public/js/script.js
+
+
+$(document).on('drop', '.js-markdown-image-upload', function(e) {
+
+  e.preventDefault();
+  var dt = e.originalEvent.dataTransfer;
```
AndrewVos: var dataTransfer

app/assets/javascripts/image-upload-markdown.js
```diff
@@ -0,0 +1,60 @@
...
+  function addImageMarkdownPlaceholder(e, filename) {
+    var startPos = e.selectionStart;
+    var endPos = e.selectionEnd;
+    e.value = e.value.substring(0, startPos)
+      + "![Uploading " + filename + "...]()\r\n"
+      + e.value.substring(endPos, e.value.length);
+  }
+
+  function replaceImageMarkdownPlaceholder(textarea, value, filename){
+    var re = new RegExp("!\\[Uploading " + filename + "\\.\\.\\.\\]\\(\\)");
```
AndrewVos: This doesn't need regex.  Normal string replace will work and be a lot less ugly.

app/assets/javascripts/image-upload-markdown.js
```diff
@@ -0,0 +1,60 @@
...
+    var request = new XMLHttpRequest();
+    request.open("POST", "/uploads");
+
+    request.onloadend = function(filename) {
+      // Form a closure around the filename.
+      return function(e) {
+        if(e.currentTarget.status == 201){
+          var filePath = e.currentTarget.response;
+          // Update the text with the correct path for the image.
+          replaceImageMarkdownPlaceholder(textarea, "!["+filename+"]("+filePath+")", filename);
```
AndrewVos: Can't the replaceImageMarkdown function define how the output ([cdsc](csdc)) looks. Seems odd having it here. WOuld be nice to just do `replaceImageMarkdownPlaceholder(textarea, file_name, file_path)`

Also consider renaming filename to fileName

app/assets/javascripts/image-upload-markdown.js
```diff
@@ -0,0 +1,60 @@
...
+    fd.append("file", file);
+
+    // CSRF token
+    fd.append("authenticity_token", $('meta[name="csrf-token"]').attr("content"));
+
+    var request = new XMLHttpRequest();
+    request.open("POST", "/uploads");
+
+    request.onloadend = function(filename) {
+      // Form a closure around the filename.
```
AndrewVos: Why do we need to do this? We're already inside a function?

AndrewVos: Also delete this comment. People reading javascript know what javascript does.

app/assets/javascripts/image-upload-markdown.js
```diff
@@ -0,0 +1,60 @@
...
+      return function(e) {
+        if(e.currentTarget.status == 201){
+          var filePath = e.currentTarget.response;
+          // Update the text with the correct path for the image.
+          replaceImageMarkdownPlaceholder(textarea, "!["+filename+"]("+filePath+")", filename);
+        } else {
+          // Remove the placeholder markup and show alert
```
AndrewVos: Delete comment

app/assets/javascripts/image-upload-markdown.js
```diff
@@ -0,0 +1,60 @@
...
+    var re = new RegExp("!\\[Uploading " + filename + "\\.\\.\\.\\]\\(\\)");
+    textarea.value = textarea.value.replace(re, function($1) {
+      return $1.replace("![Uploading " + filename + "...]()", value);
+    });
+  }
+
+  function handleFileUpload(textarea, file) {
+    addImageMarkdownPlaceholder(textarea, file.name);
+
+    var fd = new FormData();
```
AndrewVos: Rename to formData

AndrewVos:
A big thing is to consider using $textarea and passing around a jquery object.
If we're using jquery then we might as well use it.
Then you can replace all those `textarea.value = ` with jquery `$textarea.val()`.

+$(document).on('drop', '.js-markdown-image-upload', function(e) {
Or just declare ` var $textarea = $(this)` after that line ^ and then you don't have to pass it around everywhere.

AndrewVos:
Also, do the image placeholders appear next to each other when you drop more than two?
if you drop them between two paragraphs. I think you have a bug.

I would just add all the initial placeholders in a loop, and then another loop to start the file uploads.

AndrewVos:
Oh another thing, you don't really need selectionEnd unless you want to replace the text that the user has selected.

I think we should just be inserting text, not replacing anything.

fofr:
Do you think this is a feature other publishing tools might want in the future? It'd be good to use the module pattern as defined by the `govuk_admin_template`, and some of the best practices it recommends.

This will mean the JS looks and behaves in a similar way to other admin JS, with similar tests. Other apps can pull in and use the JS module, or equally it could be moved up into the admin template.

See:
https://github.com/alphagov/govuk_admin_template/blob/master/JAVASCRIPT.md
https://github.com/alphagov/govuk_admin_template/blob/master/JAVASCRIPT.md#writing-modules

fofr:
I'd expect to see some JavaScript unit tests too, preferably using Jasmine.

app/assets/javascripts/image-upload-markdown.js
tadast: > Another way might be to just upload the file in the background and immediately set the image to the one that it's going to be?

If I understand this correctly, I think we can't do that because there is no way to know what the image URL is going to be.

app/assets/javascripts/image-upload-markdown.js
tadast: We do `e.preventDefault()` on top, I think it's to trigger the event to any potential JS code that might rely on it (this was copied over from the source, URL on top of the file).

app/assets/javascripts/image-upload-markdown.js
tadast: I did it this way because I don't want to write another function that deletes the placeholder in case of an error.

app/assets/javascripts/image-upload-markdown.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  function replaceImageMarkdownPlaceholder($textarea, newValue, filename){
+    var oldPlaceholder = placeholderValue(filename);
+    $textarea.val(
+      $textarea.val().replace(oldPlaceholder, newValue)
+    );
+  }
+
+  function placeholderValue(filename) {
```
AndrewVos: There are other places you could use this function, if you changed the signature to `(filename, filepath)` ;)

app/assets/javascripts/image-upload-markdown.js
tadast: It would be nice to keep updating the placeholder with percentage values but let's see if we really need it after playing with it in production.

app/assets/javascripts/image-upload-markdown.js
tadast: We need this because multiple file uploads can happen asynchronously. There is a more detailed explanation [on the blog](http://www.stuart-nelson.co.uk/github-style-drag-n-drop-image-uploads) (cmd +f "closure")

app/assets/javascripts/image-upload-markdown.js
tadast: [What do you mean?](https://www.youtube.com/watch?v=DK_0jXPuIr0)

tadast:
@fofr this looks very good, very similar to [what I've been advocating in the past](https://coderwall.com/p/mhvucw/extended-paul-irish-s-dom-based-routing-for-coffeescript). I'll update the code and push for this style in all JS files within the application.

tadast:
@fofr would you be able to help me to kick-off the Jasmine tests or find me someone who could? I've got the basics configured but I have no idea how to set up a jasmine spec so I can test a drag-and-drop upload.

fofr:
@tadast I'd be happy to. I imagine the drag and drop could be a little fiddly.

tadast:
@fofr @AndrewVos I've reworked the PR, could you please review again?

Rakefile
```diff
@@ -4,3 +4,4 @@
 require File.expand_path('../config/application', __FILE__)
 
 Rails.application.load_tasks
+task default: "jasmine:ci"
```
AndrewVos: Does this only run the jasmine tests or does this add to the default task?

tadast: It adds, I've explained this in the commit message :)

spec/javascripts/spec/image-upload-markdown-spec.js
```diff
@@ -0,0 +1,28 @@
...
+    textarea.selectionEnd = 17;
+    // filterableTable = new GOVUKAdmin.Modules.FilterableTable();
+    // filterableTable.start(tableElement);
+  });
+
+  afterEach(function() {
+    uploadableTextarea.remove();
+  });
+
+  it('adds markdown to the current cursor position', function() {
```
AndrewVos: Nice test mate

AndrewVos:
LG

spec/javascripts/spec/image-upload-markdown-spec.js
tadast: sorry, I need to delete this file, bad rebase :dancer:

AndrewVos: If you want some help learning git just ask buddy

spec/javascripts/image-upload-markdown-spec.js
```diff
@@ -0,0 +1,64 @@
...
+  var uploadableTextarea,
+      mockDropEvent;
+
+  beforeEach(function() {
+    uploadableTextarea = $("<textarea class='js-markdown-image-upload'>First paragraph\n\nSecond paragraph</textarea>");
+
+    $('body').append(uploadableTextarea);
+    setCursorAfter("First paragraph\n");
+
+    mockDropEvent = {
```
fofr: Ahh, nicely done :)

fofr:
Good to see the Jasmine tests in, which were tricky :+1:

tadast:
@tombooth has kindly helped with the API credentials configurations so I'm going to merge <3

PR #130, opened 2016-01-11
[WIP] Add a simple topic editor

AndrewVos:


rboulton:
I thought we'd planned to create topics in collections-publisher, and also for them to have urls like `/topic/service-manual/foo`.

AndrewVos:
@rboulton The idea was to create something temporary, because we have no idea when the collections publisher features will serve our requirements. Also, @tadast and I will be working elsewhere from March.

app/controllers/topics_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def new
+    @topic = Topic.new
+  end
+
+  def create
+    topic = Topic.new(params.require(:topic).permit(:path, :title, :description))
+    topic.tree = JSON.parse(params[:topic][:tree])
+    if topic.save!
```
tadast: That's gonna raise, should be `save`, otherwise the else case is useless

app/controllers/topics_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def edit
+    @topic = Topic.find(params[:id])
+    render :new
+  end
+
+  def update
+    @topic = Topic.find(params[:id])
+    @topic.title = params[:topic][:title]
```
tadast: Should we allow updating the path until it's published?

app/controllers/topics_controller.rb
AndrewVos: Good catch!

app/views/topics/new.html.erb
```diff
@@ -0,0 +1,59 @@
+<div class="topics">
+  <% all_editions = Guide.all.map {|g| g.latest_edition} %>
```
tadast: These should be loaded in the controller

app/controllers/topics_controller.rb
AndrewVos: Yeah, but there's no code for publishing yet ;)

spec/models/topic_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'rails_helper'
+
+RSpec.describe Topic do
+  it "allows setting the topic initially" do
```
tadast: do you mean "allows setting the topic _path_"?

app/assets/javascripts/Sortable.js
```diff
@@ -0,0 +1,1265 @@
+/**!
```
tadast: this should live under vendor assets

tadast:
@AndrewVos I've spoken to Tom and it looks like we'll need to add the sidebar bit to topics
 before going live

AndrewVos:
Oh yeah totally. I think I added that to the trello ticket? Apparently we need to be able to add multiple communities?

PR #132, opened 2016-01-12
Resize textareas automatically when the content changes

AndrewVos:
Content editors find the smaller textarea to be hard to work with, so
adjust the size automatically.

### DIFF
--- a/app/assets/javascripts/text-area-autosize.js
+++ b/app/assets/javascripts/text-area-autosize.js
@@ -1,0 +1,3 @@
+$(function() {
+  $(".js-text-area-auto-size").textareaAutoSize();
+});
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -14,4 +14,5 @@
 //= require jquery_ujs
 //= require select2
 //= require protect_form_data
+//= require jquery.textarea_autosize
 //= require_tree .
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/text-area-autosize.js
```diff
@@ -0,0 +1,3 @@
+$(function() {
+  $(".js-text-area-auto-size").textareaAutoSize();
```
fofr: As I mentioned with the image uploading PR, this could be re-written slightly as a module: https://github.com/alphagov/govuk_admin_template/blob/master/JAVASCRIPT.md

Even though this is a jquery plugin, should we have some javascript unit tests?

AndrewVos: I'm not sure of the benefit of rewriting it as a module @fofr.

This is an external library which is tested here https://github.com/javierjulio/textarea-autosize/blob/master/test/test.js

app/assets/javascripts/jquery.textarea_autosize.js
```diff
@@ -0,0 +1,54 @@
+/*!
```
tadast: I think this file should live in `vendor/assets`, you'll need to explicitly require it in application.js though if I recall correctly.

tadast:
LGTM

PR #133, opened 2016-01-13
Guide slug can be changed until it has been live

AndrewVos:
https://trello.com/c/Kn6falOO/125-make-urls-modifiable-before-first-publication

### DIFF
--- a/app/models/guide.rb
+++ b/app/models/guide.rb
@@ -1,6 +1,7 @@
 class Guide < ActiveRecord::Base
   validates :content_id, presence: true, uniqueness: true
   validate :slug_format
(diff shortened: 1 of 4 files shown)


app/models/guide.rb
```diff
@@ -65,4 +70,10 @@ def slug_format
       errors.add(:slug, "can only contain letters, numbers and dashes")
     end
   end
+
+  def slug_cant_be_changed_if_an_edition_has_been_published
+    if slug != slug_was && has_published_edition?
```
tadast: `if slug_changed? &&  has_published_edition?`

PR #136, opened 2016-01-15
Remove the View Govspeak page

AndrewVos:
User testing has found that this page serves no use, and it's 90%
similar to the Edit page.

https://trello.com/c/cvNOW8Pe/144-remove-the-view-govspeak-page

screen


app/views/editions/comments.html.erb
```diff
@@ -14,7 +14,7 @@
       </a>
       <%= link_to(
             "View edition",
```
tadast: Perhaps we need to change the link label to say "View changes"/ "View differences"?

app/helpers/guide_helper.rb
```diff
@@ -16,16 +16,14 @@ def state_label(guide)
...
-      title = "Edit"
-      path =  edit_guide_path(guide)
-    elsif state == "review_requested"
-      title = "Review guide"
-      path = edition_path(guide.latest_edition)
-    elsif state == "approved"
-      title = "Publish"
-      path = edition_path(guide.latest_edition)
-    end
+    title = if state == "draft" || state == "published"
```
tadast: I think this may look a lot nicer as a case statement or even a hash
```ruby
title = { 
  "draft" => "Edit", 
  "review_requested" => "Review guide" ... 
}[state]
```

app/views/edition_changes/show.html.erb
```diff
@@ -17,7 +17,7 @@
         <br/>
         <small>
           Changes between this draft and
-          <%= link_to_if @edition.previously_published_edition, "the previous version",  @edition.previously_published_edition %>.
+          <%= link_to_if @edition.previously_published_edition, "the previous version",  edition_changes_path(@edition.previously_published_edition) %>.
```
tadast: I have a feeling this might be rather confusing to users. Perhaps it's better to remove the link and keep text-only sentence until we have reasonable edition IDs that might demystify what this link does.

tadast:
For the historic record: we're aware that this removes the ability to view past editions and replaces it with potentially inferior (for the task) "view changes", but this is not our priority at this stage and we'll revisit this topic after conducting more user research.

Other than my remaining comment this LGTM :+1:

app/views/edition_changes/show.html.erb
AndrewVos: Keep what text only sentence? "Changes between this draft and" ?

tadast: It's a sentence explaining what the user currently sees, so keep it as:
> Changes between this draft and the previous version.

But without linkifying the last part

PR #138, opened 2016-01-19
Clickable button allowing users to upload files to markdown textarea

AndrewVos:
Some users aren't able to drag+drop

https://trello.com/c/ZMllBHEJ/137-add-an-image-upload-button-as-another-way-to-initiate-image-upload

Awaiting merge of https://github.com/alphagov/service-manual-publisher/pull/129

screenshot

### DIFF
--- a/app/assets/javascripts/image-upload-markdown.js
+++ b/app/assets/javascripts/image-upload-markdown.js
@@ -1,24 +1,21 @@
 // This is a modified version of https://github.com/stuartnelson3/blog/blob/master/public/js/script.js
 
-function ImageUploadMarkdown(event){
+function ImageUploadMarkdown(files){
   var that = this;
 
   this.init = function(){
-    event.preventDefault();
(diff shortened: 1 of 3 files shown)


tadast:
Looks awesome, but I would argue the style does not indicate the area is clickable. Would it not be better to simply have a button/link? The dotted line makes me want to drag and drop files on to it rather than click it. @mikieet any thoughts?

AndrewVos:
@tadast Yeah I basically just copied [dropzone.js](http://www.dropzonejs.com/)

mikieet:
@AndrewVos Yeah I agree with @tadast, it should just be a link I reckon. Theres an image attached to the trello ticket of what I reckon it should look like. Could we do it like that but put the line of text and link before the text field (thats a finding from yesterdays research, after I made that mockup!)?

AndrewVos:
Ok moved to the top and made it a link

PR #143, opened 2016-01-21
Add pry-remote to Gemfile.

thehenster:
We normally run our dev environment with foreman/bowler so our process to be debugged is not in the foreground.


tadast:
Yay https://github.com/Mon-Ouie/pry-remote#usage

PR #145, opened 2016-01-26
Upgrade Rails to 4.2.5.1

tadast:
Looks like a part of this upgrade (we've upgraded from 4.2.4) has introduced stricter rules from the sprockets gem to explicitly include any assets referenced in views. 

It fails with an error if we don't so I've added references to favicons.


thehenster:
Looks good

PR #146, opened 2016-01-27
DO NOT MERGE - Content owner is a guide

thehenster:
https://trello.com/c/sri14201

We introduce the idea of a 'Community Guide' which is a flavour of the Guide we already have. A community guide uses the same schema (service_design_guide). Non-community guides will link to community guides to show who manages the page.

We ditch the ContentOwner and create a self referential link between guides and community guides (the relational link is actually from editions to guides).

I expect CI is bust because the schema no longer matches. We need a new version of schemas.

Todo:

* We need to publish the link to the publishing API.
* UX discussion about the checkbox which is not easy to understand.

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -125,7 +125,7 @@ private
 
     params
       .require(:guide)
-      .permit(:slug, latest_edition_attributes: [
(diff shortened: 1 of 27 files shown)


tadast:
I really like the small improvements to specs you've made everywhere :100:

PR #155, opened 2016-02-18
Ensure that guides are in a topic

AndrewVos:
@henry.turner - Pushing this up here so you can have a chance to critique this code while it's happening.

https://trello.com/c/p0l05aDj/184-published-guides-must-have-topics

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -86,12 +86,21 @@ private
   end
 
   def publish
+    unless @guide.included_in_a_topic?
+      @edition = @guide.latest_edition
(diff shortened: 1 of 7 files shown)


AndrewVos:
You mean using the `tree` attribute? When I originally wrote this topic editor I was just spiking out the UI and it ended up sticking around. It's really quite a flat structure and I would be happy to just use sql and have another table.

A bit scared of being able to do the migrations nicely in production though. Could we pair on that part tomorrow? Shouldn't be more than an hour work.

thehenster:
Go for it :+1:

PR #156, opened 2016-02-19
Publisher service

thehenster:
The overall idea was to remove the transaction-y stuff from the controller to a) clean it up a little and b) make a publisher object that would be reusable for future things that need to be published.

Just mulling.. Managing the editions is probably another area which is ready to come out of the GuidesController.. It'll be useful for Topics, Communities and Point Pages.


app/controllers/guides_controller.rb
```diff
@@ -25,17 +25,14 @@ def new
...
     @guide = Guide.new(guide_params)
 
-    ActiveRecord::Base.transaction do
-      if @guide.save
-        GuidePublisher.new(guide: @guide).put_draft
-        redirect_to edit_guide_path(@guide), notice: "Guide has been created"
-      else
-        render action: :new
-      end
+    publication = Publisher.new(content_model: @guide).
```
AndrewVos: Is that style guide stuff or should this dot be on a new line. Sorry that's incredibly pedantic

app/controllers/guides_controller.rb
```diff
@@ -25,17 +25,14 @@ def new
...
+                            save_draft(GuidePresenter.new(@guide, @guide.latest_edition))
+    if publication.success?
+      redirect_to edit_guide_path(@guide), notice: 'Guide has been created'
+    else
+      flash.now[:error] = publication.errors
```
AndrewVos: What does `.now` do?

AndrewVos:
This looks good to me, though it's going to need some big changes because I've merged that other branch in! :cry:

app/controllers/guides_controller.rb
thehenster: I can't remember seeing it in the style guide.. It's a subjective one. I've always chained at the end of the line because when copying and pasting into IRB/console it doesn't handle the dot at the beginning of the line unless it has a trailing `\`. With the dot at the end it knows there is more of the expression coming.. _shrug_ Wadda ya think?

I like pedantic :) I no-one is pedantic then code just ends up an inconsistent mess :+1:

app/controllers/guides_controller.rb
thehenster: `flash.now` is only available to the current action.. Useful for when you're not redirecting a user and rerendering a template after a failed save/update/summat.

Using `flash[:error]` without the `now` bit is the cause of those random flash messages appearing when you wouldn't expect them if you've seem them lurking :ghost:

thehenster:
Cool. I had a thought about the stuff thats gone into the guides controller to stop the user publishing without a topic. I saw that we have states on the editions.. draft, published.. and one other.. can't remember. We could keep the same `#included_in_a_topic?` but swap the controller bit for a validation that dips into the latest edition and checks it's state?

Keeps the business logic a bit more in the model and would make merging this PR a bit less conflict-y?

app/controllers/guides_controller.rb
AndrewVos: cool

PR #159, opened 2016-02-23
Store Guide id in topic tree instead of edition id

AndrewVos:
This allows us to publish topics without having to update editions in the
topic to new versions.


thehenster:
LGTM - I can't see any other "editions" in there..

AndrewVos:
We just have to be sure that it's not going to make the UI unusable in production.

PR #168, opened 2016-02-25
Community as a type of guide again

thehenster:
A guide's community was represented by a ContentOwner was a hardcoded thing. We only had two ContentOwners to choose from.

Here we change the way we reference the guide's community. Because the a guide and a guide community are so similar we introduce an STI to give use Guide and GuideCommunity where a Guide has a (self referential relation) to GuideCommunity.

This is a reincarnation of https://github.com/alphagov/service-manual-publisher/pull/146 where there is a little discussion. A fair few of the commits are the same.

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -18,7 +18,9 @@ class GuidesController < ApplicationController
   end
 
   def new
-    @guide = Guide.new(slug: "/service-manual/")
+    type = params[:community].present? ? 'GuideCommunity' : nil
+
+    @guide = Guide.new(slug: "/service-manual/", type: type)
     @guide.build_latest_edition(update_type: 'major')
   end
 
@@ -126,7 +128,7 @@ private
 
     params
       .require(:guide)
-      .permit(:slug, latest_edition_attributes: [
+      .permit(:slug, :type, latest_edition_attributes: [
         :title,
         :body,
         :description,
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -2,9 +2,9 @@ class GuidesController < ApplicationController
   def index
     @user_options = User.pluck(:name, :id)
     @state_options = %w(draft published review_requested approved).map { |s| [s.titleize, s] }
-    @content_owner_options = ContentOwner.pluck(:title, :id)
 
-    @guides = Guide.includes(latest_edition: [:user, :content_owner])
+    # TODO: :content_owner not being included is resulting in an N+1 query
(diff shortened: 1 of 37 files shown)


app/controllers/guides_controller.rb
```diff
@@ -2,9 +2,9 @@ class GuidesController < ApplicationController
   def index
     @user_options = User.pluck(:name, :id)
     @state_options = %w(draft published review_requested approved).map { |s| [s.titleize, s] }
-    @content_owner_options = ContentOwner.pluck(:title, :id)
 
-    @guides = Guide.includes(latest_edition: [:user, :content_owner])
+    # TODO: :content_owner not being included is resulting in an N+1 query
```
AndrewVos: Can you fix this todo?

db/migrate/20160225101236_remove_content_owner_not_null_constraint_from_editions.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveContentOwnerNotNullConstraintFromEditions < ActiveRecord::Migration
+  def change
+    change_column :editions, :content_owner_id, :integer, null: true
```
AndrewVos: http://apidock.com/rails/ActiveRecord/ConnectionAdapters/SchemaStatements/change_column_null

db/migrate/20160225130400_drop_content_owners.rb
```diff
@@ -0,0 +1,61 @@
+class DropContentOwners < ActiveRecord::Migration
```
AndrewVos: This is pretty complex :( Couldn't it be done in sql?

thehenster: It'd be worse I think. It's got if conditions in it. It was never going to be a thing of beauty..

db/migrate/20160225101236_remove_content_owner_not_null_constraint_from_editions.rb
thehenster: Noice - I didn't know about that. Thanks

app/controllers/guides_controller.rb
thehenster: I had a quick crack but it's not a quick fix. The problem is that because the :content_owner is a self referential join when searching we add a condition on the `tsv` column which is ambiguous because it's hand crafted SQL.

I'm up for fixing too but would prefer a separate PR?

AndrewVos:
Other than that one migration comment this lgtm. :ship:

PR #175, opened 2016-03-01
Validate links when publishing guides

AndrewVos:
- Ensure links in guide body are 200. Allow user to publish even if there
are broken links.
- Cache link checking, but expire it after five minutes.

## Guide couldn't be published because it had broken links:
image

## But, we can just publish anyway:
image


thehenster:
I like the UI and the option to force publish. The validation errors are a good place to display the errors too.

Could we show the screenshots to the content lot to see what they think?

The worry is that a page with a lot of links could take a long time to save. I think that's worth mentioning to the content people too. If they're pressing that save button a lot over a 2 hour period then the slow down could be too much for them.

I'll have a look at the code..

AndrewVos:
@thehenster the links are only checked when Publish is clicked. This is because I was also thinking that the user would be clicking Save more often. It shouldn't be too slow, though perhaps we could do what `whitehall` does and use Typhoeus for parallel http requests.

thehenster:
Looks good although maybe we could show this to everyone before merging? It's quite a major bit of publishing flow.

AndrewVos:
Reminder: do a demo tomorrow

AndrewVos:
This isn't going to work on environments with http basic auth :cry:

AndrewVos:
Maybe this should be done entirely in javascript.

thehenster:
LGTM

PR #178, opened 2016-03-02
Fix comments form URL for a guide community

thehenster:
Because Guide is a now an STI the polymorphic URL generation is based of the model_name whereas we specifically want it to use the same route and controller.


AndrewVos:
lgtm

PR #179, opened 2016-03-03
Switch to using factory_girl for generating test data

AndrewVos:


Gemfile
```diff
@@ -47,4 +47,5 @@ group :test do
   gem 'launchy'
   gem 'poltergeist'
   gem 'database_cleaner'
+  gem 'factory_girl_rails', '~> 4.0'
```
thehenster: We need to pin the version? Can't bundler decide?

spec/controllers/comments_controller_spec.rb
```diff
@@ -4,9 +4,9 @@
 
   let(:commenter) { User.create(name: "Commenter", permissions: ['signin'], email: "commenter@example.com") }
   let(:edition) do
-    edition = Generators.valid_edition(guide: Guide.new(slug: "/service-manual/commentable"))
-    edition.save!
-    edition
+    guide = build(:guide, slug: "/service-manual/commentable")
+    guide.save!
```
thehenster: Use a `.create`?

spec/factories.rb
```diff
@@ -0,0 +1,67 @@
+FactoryGirl.define do
+  factory :guide_community, class: GuideCommunity do
```
thehenster: Doesn't it guess the class for you if it has a `classify`able name?

Gemfile
AndrewVos: Yeah apparently we do, according to the readme.

thehenster: I'm not calling you a liar :) .. sometimes pinning has to happen to avoid conflicts.. but where does it say it in the readme? https://github.com/thoughtbot/factory_girl_rails

AndrewVos: http://www.rubydoc.info/gems/factory_girl/file/GETTING_STARTED.md

AndrewVos: Works without pinning so removing it

thehenster: :+1:

spec/models/edition_spec.rb
```diff
@@ -33,42 +34,45 @@
...
+      @edition2 = create(:published_edition, guide: @guide)
+      @edition3 = create(:published_edition, guide: @guide)
+      @edition4 = create(:published_edition, guide: @guide)
+      @guide.editions = [
+        @edition1,
+        @edition2,
+        @edition3,
+        @edition4,
+      ]
+    end
```
thehenster: I like this refactor btw. More readable :+1:

thehenster:
:clap: Looks good!

PR #180, opened 2016-03-03
Upgrade gds-api-adapters

AndrewVos:
Rename #put_links to #patch_links because it has been changed in the
client library.


thehenster:
:+1:

PR #181, opened 2016-03-03
Use FactoryGirl#build_stubbed/build where possible, to speed up tests

AndrewVos:


thehenster:
I've put a couple of comments here.

I've never really used .build_stubbed. If I'm honest, it scares me, but I'm keen to try it.

PR #182, opened 2016-03-07
Topic drafts

thehenster:
This let's us save drafts to the content store and publish them separately.

This doesn't let us see previous versions and compare them. In fact, we are still just updating the same model and have not implemented versions/editions.

Because we aren't saving versions/editions and their state, the system doesn't know whether a topic has been published yet. For the moment this is down to the user. Whoever is writing the content will know whether they have pressed the publish button yet. For everyone else they will have to compare the preview and live versions.

One day we will add versioning/editions but it's to be prioritised.


AndrewVos:
lgtm

PR #183, opened 2016-03-08
Remove duplicate guides from the topic admin dropdown

thehenster:
The joined editions were causing duplicate guides. Rails's two queries generated by #includes was pretty neat in the first place. A subquery would be required to make a single database query which really isn't worth it.

### DIFF
--- a/app/helpers/topic_helper.rb
+++ b/app/helpers/topic_helper.rb
@@ -5,8 +5,8 @@ module TopicHelper
 
   def all_guides_container_for_select
     @_all_guides_container_for_select ||=
-      Guide.includes(:latest_edition).
-            order('editions.title').
-            pluck('editions.title', 'guides.id')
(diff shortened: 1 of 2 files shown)


app/helpers/topic_helper.rb
```diff
@@ -5,8 +5,8 @@ def view_topic_url(topic)
 
   def all_guides_container_for_select
     @_all_guides_container_for_select ||=
-      Guide.includes(:latest_edition).
-            order('editions.title').
-            pluck('editions.title', 'guides.id')
+      Guide.includes(:latest_edition).sort_by(&:title).map do |guide|
```
AndrewVos: Do we still need this `includes`?

AndrewVos: Oh wait yes we need it for the delegated `Guide#title`

PR #187, opened 2016-03-08
Set up GdsApi logging

thehenster:
https://github.com/alphagov/gds-api-adapters/blob/master/README.md#logging

### DIFF
--- a/config/initializers/gds_api.rb
+++ b/config/initializers/gds_api.rb
@@ -1,0 +1,3 @@
+require 'gds_api/base'
+
+GdsApi::Base.logger = Logger.new(Rails.root.join("log/#{Rails.env}.api_client.log"))


config/initializers/gds_api.rb
```diff
@@ -0,0 +1,3 @@
+require 'gds_api/base'
+
+GdsApi::Base.logger = Logger.new(Rails.root.join("log/#{Rails.env}.api_client.log"))
```
AndrewVos: Will this log to kibana or? Not sure what happens here

thehenster: Just a good ole text file.

The JSON logs aren't getting pulled into Kibana for this app yet anyway so no rush either :)

AndrewVos: Oh ok but what I mean is how will we see these logs? Is this just gonna be used in development? I'm scared that we unwittingly use up all the disk space on some production server. Logs get big quiiiick:
```
[~/.../service-manual-publisher] master ∞ du -h --max-depth 1
544K	./coverage
180K	./db
20K	./bin
32M	./.git
16K	./lib
53M	./tmp
292K	./spec
20K	./vendor
918M	./log
108K	./config
20K	./public
456K	./app
1004M	.

```

AndrewVos:
Maybe just add a conditional `if Rails.env.development?`

thehenster:
All the hosted environments run with the production rails env. I really want the logging on integration and staging.

To change the behaviour in the different hosted environments we'd need to overwrite this file with a template from puppet.

I _think_ that all files in the log directory are rotated anyway.. Just looking in the other apps I can't see any log files that aren't..

PR #189, opened 2016-03-11
Ensure guide paths always include the topic name

AndrewVos:
For now this is to be done manually.

### DIFF
--- a/app/models/guide.rb
+++ b/app/models/guide.rb
@@ -84,6 +84,8 @@ private
       errors.add(:slug, "must be filled in")
     elsif !slug.to_s.match(/\A\/service-manual\/[a-z0-9\-\/]+$/i)
       errors.add(:slug, "can only contain letters, numbers and dashes")
+    elsif !slug.to_s.match(/\A\/service-manual\/[a-z0-9-]+\/[a-z0-9-]+/)
+      errors.add(:slug, "must be present and start with '/service-manual/[topic]'")
     end
   end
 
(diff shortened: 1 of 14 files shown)


spec/features/slug_migration_spec.rb
```diff
@@ -146,7 +146,7 @@ def create_slug_migration_without_redirect_to slug
   end
 
   it "migrates to a topic path" do
-    create(:topic, path: "/service-manual/topic-1")
+    create(:topic, path: "/service-manual/topic-name/topic-1")
```
thehenster: Topic nested under a topic?

AndrewVos:
@thehenster fixed

app/models/guide.rb
```diff
@@ -84,6 +84,8 @@ def slug_format
       errors.add(:slug, "must be filled in")
     elsif !slug.to_s.match(/\A\/service-manual\/[a-z0-9\-\/]+$/i)
       errors.add(:slug, "can only contain letters, numbers and dashes")
+    elsif !slug.to_s.match(/\A\/service-manual\/[a-z0-9-]+\/[a-z0-9-]+/)
+      errors.add(:slug, "must be present and start with '/service-manual/[TOPIC]'")
```
thehenster: I reckon this is the right approach for the moment.

We could check for the presence of the actual topic name _or even_ generate the whole prefix and only let the user edit the last part of the slug. The user can change the slug right up until publishing. Too much at the moment though I reckon.

You could use curly brackets for repetition in that regex. eg. `/\A\/service-manual(\/[a-z0-9-]+){1,2}`

AndrewVos: Feels a bit more complex to me maybe?

PR #190, opened 2016-03-11
Use search index mainstream with document type service_manual

thehenster:
So that we don't share the same index as the original service manual, which is purged upon each deploy, we are going to use the mainstream index and a specific document type so that search queries can downgrade service manual results (as per current functionality).

Here is the related PR in rummager for the new document type and the downgrading bit: https://github.com/alphagov/rummager/pull/586

### DIFF
--- a/app/models/search_indexer.rb
+++ b/app/models/search_indexer.rb
@@ -6,10 +6,10 @@ class SearchIndexer
 
   def index
     index = Rummageable::Index.new(
-      Plek.current.find('rummager'), '/service-manual'
+      Plek.current.find('rummager'), '/mainstream'
     )
     index.add_batch([{
-      "_type":             "manual_section",
+      "_type":             "service_manual",
       "description":       @edition.description,
       "indexable_content": @edition.body,
       "title":             @edition.title,
--- a/app/models/search_indexer.rb
+++ b/app/models/search_indexer.rb
@@ -9,6 +9,7 @@ class SearchIndexer
       Plek.current.find('rummager'), '/mainstream'
     )
     index.add_batch([{
+      "format":            "service_manual",
       "_type":             "service_manual",
       "description":       @edition.description,
       "indexable_content": @edition.body,
--- a/app/models/topic_search_indexer.rb
+++ b/app/models/topic_search_indexer.rb
@@ -5,6 +5,7 @@ class TopicSearchIndexer
 
   def index
     rummager_index.add_batch([{
+      "format":            "service_manual",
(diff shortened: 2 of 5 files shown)


thehenster:
Updated to always using the format "service_manual" and the _type "service_manual".

rboulton:
Is there a difference in the data stored in a "service_manual" document, compared to a "manual_section" document?  I'm assuming not, since you're not changing the data sent along with this change.  It's desirable not to unneccesarily increase the number of formats which exist, since that makes rummager's job more complex, and makes it harder to do things like changing the way all manuals work.

If you _do_ decide to introduce a new format, it will.  I also think it should be named "service_manual_section", since it's not the whole manual.  (For the existing modelling, there's a "manual" format, which multiple "manual_sections" are then part of.)

If you're just wanting to change the weight of documents in the service manual, that could be accomplished by looking at the value of the "manual" field for each document, and downweighting ones which have the value "service-manual".  This would be a more generic and extensible approach for the future (all manuals may need weighting according to the importance of the manuals).  This may also not be required, since the popularity weighting does this type of effect for all formats.

AndrewVos:
@rboulton Does the "manual" type exist on mainstream?

rboulton:
Yes - for example: https://www.gov.uk/api/search.json?filter_link=/guidance/immigration-rules

(Just had a face-to-face chat with @tijmenb who comments, reasonably, that keeping the downweighting code simple is much easier if you introduce a new format for service manual documents, so that might well be the way to go.  I'd still suggest using "service_manual_section" as the format name instead of "service_manual", though.)

AndrewVos:
I think a name that applies to our project might be `service_manual_guide`?

AndrewVos:
Actually, does this include topics too? Maybe `service_manual_section` is the one!

thehenster:
If we do go with the format `service_manual_section` should we also create `service_manual` which would have a single document to represent the whole service manual? If so should we down weight this single document's format `service_manual` too? Personally, I don't think so. The service manual itself seems sufficiently important.

Also, I think I overlooked adding the new format(s) to https://github.com/alphagov/rummager/blob/master/config/schema/indexes/mainstream.json

thehenster:
I don't know the query language for elastic search. But if we were to consider the code changes to down weight via the "manual" value. Would this do it? https://github.com/alphagov/rummager/commit/dd0c2186c514e19b0f3574a7612da47625108a1c

thehenster:
Closing in favour of https://github.com/alphagov/service-manual-publisher/pull/191

PR #191, opened 2016-03-15
Use mainstream search index

thehenster:
So that we don't share the same index as the original service manual, which is purged upon each deploy, we are going to use the mainstream index and a specific document type so that search queries can downgrade service manual results (as per current functionality).

Here is the related PR in rummager for the new document type and the downgrading bit: alphagov/rummager#589

This is a reincarnation of https://github.com/alphagov/service-manual-publisher/pull/190


AndrewVos:
lgtm

PR #192, opened 2016-03-17
Validate urls in parallel

AndrewVos:
We are getting timeouts in integration if all the urls are requested
synchronously.


app/models/govspeak_url_checker.rb
```diff
@@ -4,15 +4,22 @@ def initialize govspeak
...
+    end
 
-      !checked_url.ok?
+    broken_urls = Parallel.map(urls_to_check, in_threads: 10) do |u|
+      check_url(u) ? nil : u
+    end.compact
+
+    urls_to_check.each do |url|
+      CheckedUrl.find_or_initialize_by(url: url) do |checked_url|
+        checked_url.ok = !broken_urls.include?(checked_url.url)
```
benlovell: You could use `exclude?` to avoid the negation.

app/models/govspeak_url_checker.rb
```diff
@@ -4,15 +4,22 @@ def initialize govspeak
...
-        checked_url.save!
-      end
+    urls_to_check = extract_all_urls_from_govspeak.select do |url|
+      url = CheckedUrl.find_or_initialize_by(url: url)
+      url.new_record? || url.expired?
+    end
 
-      !checked_url.ok?
+    broken_urls = Parallel.map(urls_to_check, in_threads: 10) do |u|
+      check_url(u) ? nil : u
```
benlovell: Did you first try changing `#check_url` to perform a `HEAD` request? This is probably piles more efficient.

app/models/govspeak_url_checker.rb
```diff
@@ -4,44 +4,63 @@ def initialize govspeak
...
+  def check_urls urls
+    broken_urls = []
+    hydra = Typhoeus::Hydra.new(max_concurrency: 5)
+
+    urls.each do |url|
       options = {
-        follow_redirects: true,
-        timeout: 5,
+        followlocation: true,
+        forbid_reuse: true,
```
thehenster: Does this do anything when each request is in a separate thread? I couldn't imagine it managing to keep persistent connections across threads for some reason..

app/models/govspeak_url_checker.rb
```diff
@@ -4,44 +4,63 @@ def initialize govspeak
...
       end
 
       begin
-        HTTParty.get(
-          url,
-          options,
-        ).ok?
-      rescue
-        false
+        Addressable::URI.parse(url)
```
thehenster: Could we add addressable to the gemfile if we're gonna use it over normal URI? It looks like it's just a fortunate dependency of other gems atm.

app/models/govspeak_url_checker.rb
```diff
@@ -4,44 +4,63 @@ def initialize govspeak
...
+      url = CheckedUrl.find_or_initialize_by(url: url)
+      url.new_record? || url.expired?
+    end
 
-      !checked_url.ok?
+    if urls_to_check.any?
+      broken_urls = check_urls(urls_to_check)
+      urls_to_check.each do |url|
+        CheckedUrl.find_or_initialize_by(url: url) do |checked_url|
+          checked_url.ok = !broken_urls.include?(checked_url.url)
```
thehenster: Ben's `exclude?` suggestion would be nice.

app/models/govspeak_url_checker.rb
```diff
@@ -4,44 +4,63 @@ def initialize govspeak
...
+        end.save!
```
thehenster: If saving immediately we could use http://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-find_or_create_by-21 ?

app/models/govspeak_url_checker.rb
```diff
@@ -4,44 +4,63 @@ def initialize govspeak
...
+      []
     end
   end
 
   private
 
-    def check_url url
+  def check_urls urls
+    broken_urls = []
+    hydra = Typhoeus::Hydra.new(max_concurrency: 5)
```
thehenster: Does it spew the same error regarding AR connections if we go above 5?

app/models/govspeak_url_checker.rb
AndrewVos: Yes it's curl that handles the parallel processing afaik

app/models/govspeak_url_checker.rb
AndrewVos: Done

PR #193, opened 2016-03-18
Fix issue where Guide Communities could not be created

AndrewVos:
Allow url paths without a topic name in them


AndrewVos:
Just occurred to me that this is probably incorrect because these paths look like `/communities/community-name`. Will rebase this

PR #196, opened 2016-03-21
Ignore old content that can't be seeded

AndrewVos:
Some content has a slug format that is not compatible with our new
approach to slugs

For example `/service-manual/agile-delivery` will not work, because we
require a topic name like `/service-manual/agile/agile-delivery` to be
present.

Also, move the seed task out of seeds into its own task, because it's not really seed data and is only used in development.


lib/tasks/import_old_service_manual_content.rake
```diff
@@ -0,0 +1,102 @@
...
+        body: body,
+      }
+    end
+  end
+
+end
+
+desc "import old content from government-service-design-manual"
+task import_old_service_manual_content: :environment do
+  author = User.first || User.create!(name: "Unknown", email: "unknown@example.com")
```
thehenster: Isn't this the default user of the system? Don't we need this to be able to log into the app for the first time after a fresh db:reset?

AndrewVos: Done.

PR #197, opened 2016-03-21
Auto-link text in comments

AndrewVos:
image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,7 @@ gem 'auto_strip_attributes', '~> 2.0', '>= 2.0.6'
 gem 'httparty', '~> 0.13.7'
 gem 'addressable', '~> 2.3.7'
 gem 'typhoeus', '~> 1.0', '>= 1.0.1'
+gem 'rinku', '~> 1.7', '>= 1.7.3', require: "rails_rinku"
 
 group :development do
   gem 'better_errors'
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -26,6 +26,7 @@ gem 'auto_strip_attributes', '~> 2.0', '>= 2.0.6'
 gem 'httparty', '~> 0.13.7'
 gem 'addressable', '~> 2.3.7'
 gem 'typhoeus', '~> 1.0', '>= 1.0.1'
+gem 'rinku', '~> 1.7', '>= 1.7.3', require: "rails_rinku"
```
thehenster: What's with all the gem pinning?

spec/features/comments_spec.rb
```diff
@@ -12,18 +12,35 @@
...
+
+    it "auto links urls in comments" do
+      guide = create(:guide)
+      write_a_comment_successfully(
+        guide: guide,
+        comment: "This is a link: http://google.com",
+      )
+      within ".comments .comment" do
+        expect(page).to have_link "http://google.com", href: "http://google.com"
+      end
```
thehenster: Nicely written test.

Maybe we could delete the one above? It seems to be covered by this one.

Gemfile
AndrewVos: This particular gem only patches rails `auto_link` method if you require `rails_rinku`. It literally has no other use to us, so I added it in here.

spec/features/comments_spec.rb
AndrewVos: I agree with you but not 100% sure. I'll look at the code again and have a think.

AndrewVos: refactored.

PR #198, opened 2016-03-22
Use service manual frontend

thehenster:
Only when we are ready to start republishing to the new frontend should this be merged and deployed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ PostgreSQL-backed Rails 4 "Publishing 2.0" application for internal use, with no
 **To persist and render guides**
 
 - [content-store](https://github.com/alphagov/content-store)
-- [government-frontend](https://github.com/alphagov/government-frontend)
+- [service-manual-frontend](https://github.com/alphagov/service-manual-frontend)
 
 **To index and search published guides**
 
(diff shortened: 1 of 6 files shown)


dougdroper:
:+1:

thehenster:
Hey @AndrewVos. I deployed this branch to integration and it seems to be behaving. Merging this will stop us deploying the publisher to production until the new frontend is in production.. But it should be this week. Merge or hold off?

AndrewVos:
Why would it stop us deploying the new publisher @thehenster?

thehenster:
This changes the rendering_app to be service-manual-frontend. If we deploy the publisher, publish something and try and view it in the frontend it won't be able to find the rendering app service-manual-frontend.

As we're on the home straight with the frontend I think it's worth it so we can get more integration testing under our belt.. but asked in case you might have something you're keen to get out?

AndrewVos:
No I don't have anything pressing. Are we really ready to release the new frontend though? I think that should be deployed before this gets merged.

thehenster:
The only benefit is a bit of extra testing in integration. No bother at all though.. let's leave it.

PR #199, opened 2016-03-22
Remove URL validation feature

AndrewVos:
We are going to icebox this because it's hard to design a good
experience for this feature. Instead, the Custom team are building an
app that handles this sort of thing. We'll use that API when it's
available.


thehenster:
:+1:

PR #204, opened 2016-03-22
Fix bug where we were not sending topics to the correct content items

AndrewVos:
~~Don't merge, because this needs tests and need to fix the bug in the `publishing-api` if possible.~~
Also, turns out if you pass an incorrect ID through to `patch_links`,
you get a 200.

### DIFF
--- a/spec/models/guide_tagger_job_spec.rb
+++ b/spec/models/guide_tagger_job_spec.rb
@@ -1,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe GuideTaggerJob do
+  let(:api_double) { double(:publishing_api) }
+
+  before do
+    stub_const("PUBLISHING_API", api_double)
+  end
+
+  describe "#batch_perform_later" do
+    it "adds the topic to all guides" do
+      topic = create(:topic, :with_some_guides)
+      guide_ids = topic.tree.map {|t| t["guides"]}.flatten
+      guide_content_ids = Guide.where(id: guide_ids).pluck(:content_id)
+      guide_content_ids.each do |guide_content_id|
+        expect(api_double).to receive(:patch_links)
+          .with(
+            guide_content_id,
+            links: { topics: [topic.content_id] },
+          )
+      end
+
+      GuideTaggerJob.batch_perform_later(topic)
+    end
+  end
+end
--- a/spec/models/topic_spec.rb
+++ b/spec/models/topic_spec.rb
@@ -49,11 +49,14 @@ RSpec.describe Topic do
     end
   end
 
-  describe "#guide_ids" do
-    it "returns the associated guide ids" do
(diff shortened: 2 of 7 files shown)


thehenster:
I'd be up for changing that `topic_id:` parameter too.. It should be `topic_content_id:` in the class and instance method?

I think it was me that introduced this bug but I'm blaming the naming :)

AndrewVos:
Yeah, not sure why the tests didn't fail when I opened this PR :/

AndrewVos:
Definitely not ready for merge. Need to fix the publishing API bug

thehenster:
I think as it stands it fixes the problem no? My suggestion was just for readability.

AndrewVos:
Added a test and refactored. Merge pls

dougdroper:
:+1:

spec/models/guide_tagger_job_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+          .with(
+            guide_content_id,
+            links: { topics: [topic.content_id] },
+          )
+      end
+
+      GuideTaggerJob.batch_perform_later(topic)
+    end
+  end
+end
```
thehenster: Good work adding this :clap:

AndrewVos: Thanks mate, I'll teach you about TDD someday ;)

AndrewVos: OOooh never mind I checked `git blame` and I'm gonna have to retract my burn. Nice work @tadast

spec/models/topic_spec.rb
```diff
@@ -49,11 +49,13 @@
...
-      topic = create(:topic, tree: [{'guides' => ['2', '5']}, {'guides' => ['3']}])
-
-      expect(topic.guide_ids).to match_array([2, 3, 5])
+  describe "#guide_content_ids" do
+    it "returns the associated guide content ids" do
+      topic = create(:topic, :with_some_guides)
+      guide_ids = topic.tree.map {|t| t["guides"]}.flatten
+      guide_content_ids = Guide.where(id: guide_ids).pluck(:content_id)
+      expect(topic.guide_content_ids).to match_array(guide_content_ids)
+      expect(topic.guide_content_ids).to_not be_empty
```
thehenster: I find this less readable than the original but I'm not too sure how to improve it :( The content ids are generated so I guess we'll never get that nice readability of seeing the literals in the setup and the assertions.. Oh well.

Maybe stick a line break above the `expect`s? That'll fix it ;)

thehenster:
Defo better passing the topic rather than the arguments for the publishing-api call.

spec/models/topic_spec.rb
AndrewVos: I could set the `tree` attribute in the test I suppose. :(

PR #205, opened 2016-03-23
Display logical numbers on editions in history

AndrewVos:


thehenster:
Overall I'd be tempted to add this number to the database. We want the numbers to be consistent as people are going to reference the numbers in discussion. And when you start throwing in deleting drafts and pagination I think it'll end up simpler with it in the DB?

But just regarding this code.. I'd prefer it being done more in the view layer instead of the virtual attribute and controller approach.

AndrewVos:
@thehenster Yup just what doug was saying. Have a look again ;)

thehenster:
We'll need to revist if we ever start paginating or deleting drafts these but I guess we can worry about that then.

PR #207, opened 2016-03-23
Change search filter labels to something more obvious

AndrewVos:
https://trello.com/c/Jfv3CG9g/283-fix-labels-on-filters-and-table-headers-in-pub-app


thehenster:
Looks good.

Is there a separate story for changing the label of the "Published By" dropdown on the guide edit page?

AndrewVos:
Don't think there is

thehenster:
I reckon we should do it as part of the same story and we should write the story and stick it at the top of 'Next'. I think without it we're kind of making it worse..

PR #208, opened 2016-03-23
Fix N+1

AndrewVos:


thehenster:
I'm not sure but I have a feeling fixing this isn't so simple.. content owners are a self referential join. Won't the generated query be searching across the guide and it's related content owner?

AndrewVos:
Not sure what you mean?

thehenster:
That query is joining guides to guide communities and those models use STI to live in the same `guides` table so it's joining `guides` to `guides`.

Say you have a guide called "Scrum" and it's guide community is called "Agile". If you so a search for "Agile" it's going to return you the scrum guide because the query is ambiguous about whether you're talking about the guides or guide communities?

PR #209, opened 2016-03-29
Store change history for each published guide with update_type=major

AndrewVos:
https://trello.com/c/qzHbv9S9/181-show-change-notes-and-page-history-at-the-bottom-of-guidance-page

Dependent on:

https://github.com/alphagov/government-frontend/pull/129
https://github.com/alphagov/service-manual-publisher/pull/209
https://github.com/alphagov/govuk-content-schemas/pull/274

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -147,8 +147,8 @@ private
         :related_discussion_href,
         :related_discussion_title,
         :update_type,
+        :reason_for_change,
         :change_note,
-        :change_summary,
       ]).deep_merge(with)
   end
 
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -12,13 +12,14 @@ class Edition < ActiveRecord::Base
 
   scope :draft, -> { where(state: 'draft') }
   scope :published, -> { where(state: 'published') }
+  scope :major_published_editions, -> { published.where(update_type: 'major')}
   scope :review_requested, -> { where(state: 'review_requested') }
   scope :most_recent_first, -> { order('created_at DESC, id DESC') }
 
   validates_presence_of [:state, :phase, :description, :title, :update_type, :body, :user]
   validates_inclusion_of :state, in: STATES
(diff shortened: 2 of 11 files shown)


app/controllers/guides_controller.rb
```diff
@@ -147,8 +147,8 @@ def guide_params(with = {})
         :related_discussion_href,
         :related_discussion_title,
         :update_type,
+        :change_reason_for_change,
```
thehenster: Ha :)

Maybe it'd be easier if these columns didn't match the schema fields. `change_summary` and `change_description` and we map them in the presenter?

app/models/edition.rb
```diff
@@ -88,6 +89,7 @@ def change_note_html
   def draft_copy
     dup.tap do |e|
       e.change_note = nil
+      e.change_reason_for_change = nil
```
thehenster: I think this method along with #ensure_draft_exists can go can't it? They're no longer used?

app/controllers/guides_controller.rb
thehenster: Or just name it `editons.reason_for_change`? The `change_reason_for_change` is a bit of a mouthful.

app/models/edition.rb
AndrewVos: Not sure?

app/controllers/guides_controller.rb
AndrewVos: K will go with the latter

app/models/edition.rb
thehenster: They're no longer used. You can delete `Guide#ensure_draft_exists` and `Edition#draft_copy` along with tests for both.

AndrewVos: Out of the scope of this change, surely?

thehenster: I'm a liar. `Edition#draft_copy` is used in `Guide#latest_editable_edition`. I'll delete the other method separately.

thehenster:
Looks good to me

PR #211, opened 2016-03-31
Stop redirecting back to root after modifying a guide

AndrewVos:
https://trello.com/c/ZYB8fHOV/288-save-from-publisher-should-keep-user-on-the-edit-guide-page-instead-of-dashboard


thehenster:
Looks good

PR #214, opened 2016-04-04
Some UI changes

AndrewVos:


thehenster:
That column title should really be "Last updated by" until we do the story that implements assigning editions to people. But it's not worse.. so..

LGTM

PR #216, opened 2016-04-06
Remove the word Community from the guide index community titles

AndrewVos:
https://trello.com/c/EjzqwXL6/293-quick-fixes-to-lables-and-inputs-on-the-guide-list-table

### DIFF
--- a/spec/features/guide_index_spec.rb
+++ b/spec/features/guide_index_spec.rb
@@ -1,0 +1,19 @@
+require 'rails_helper'
+
+RSpec.describe 'Guide index', type: :feature do
+  it "removes the Community part from guide community titles" do
+    guide = create(:guide)
+
+    visit root_path
+
+    content_owner_title =
+      guide.latest_edition.content_owner.title
+    content_owner_title_without_community =
+      guide.latest_edition.content_owner.title.gsub(" Community", "")
+
+    within_guide_index_row guide.title do
+      expect(page).to_not have_content content_owner_title
+      expect(page).to have_content content_owner_title_without_community
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


app/models/guide.rb
```diff
@@ -59,6 +59,10 @@ def ensure_draft_exists
     self
   end
 
+  def title_without_community
+    self.title.gsub(/ [cC]ommunity$/, "")
+  end
```
thehenster: I still wouldn't have this at the model layer. It's more of a helper to me.

spec/features/guide_index_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    expected_row = [
+      anything,
+      guide.latest_edition.content_owner.title.gsub(" Community", ""),
+      anything,
+      anything,
+      anything,
+    ]
+    expect(rows).to include expected_row
+  end
+end
```
thehenster: We have a helper method which might make for a nicer test. https://github.com/alphagov/service-manual-publisher/blob/master/spec/support/capybara_helpers.rb#L6-L8

Something like the following would be less susceptible to future changes to the table.

```
within_guide_index_row("Scrum") do
  expect(page).to have_css(".community", "Agile Delivery")
end
```

There are other options too obv..

app/helpers/application_helper.rb
```diff
@@ -20,4 +20,8 @@ def css_class_if_current(path, class_name)
       class_name
     end
   end
+
+  def guide_community_title(guide_community)
+    guide_community.title.gsub(/ [cC]ommunity$/, "")
+  end
```
thehenster: I reckon this could move to the guide_helper.rb?

thehenster:
Looks good :+1:

PR #218, opened 2016-04-06
Fix broken compare page for communities

AndrewVos:


AndrewVos:
This is nice, I like it.

PR #219, opened 2016-04-07
Fix broken tests and refactor edition events

AndrewVos:


spec/features/guide_history_spec.rb
```diff
@@ -1,9 +1,18 @@
...
+
+  after do
+    travel_back
+  end
+
   it "shows a history of the latest edition" do
     stub_publisher
     create_guide_community
+
+    travel_to Time.new(2004, 11, 24, 01, 04, 44)
```
thehenster: This guy can take a block which saves using the `after` callback. I think it reads nicer using the block form because you don't have to look outside the `it` block when trying to work out what's going on..

AndrewVos: Nice didn't know that

app/views/editions/edition_thread/_assigned_to_event.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class="event panel panel-default">
   <div class="panel-heading">
     <div class="pull-right">
-      <%= format_time(event.try(:updated_at) || event.try(:at)) %>
+      <%= format_time(event.edition.updated_at) %>
```
thehenster: If not `event.edition.updated_at.to_date.to_s(:rfc822)` then you can add a new format.. Doing it via internationalisation is the trendiness I think.

http://guides.rubyonrails.org/i18n.html#adding-date-time-formats

AndrewVos: I've written this code with the idea that it will probably change soon and just use a built in time format.

thehenster:
Looks good.

PR #222, opened 2016-04-08
[REVERT] Store change history for each published guide with update_type=major

AndrewVos:
Turns out that the latest published item can never get added to this change
history, because it is never published when we're presenting this data in
`put_content`.

We need to do one of these:

1. Store `note` and `reason_for_change` on the guide content, and add it to change history in the frontend.
2. Redo `put_content` every time we publish.
3. ???


thehenster:
Good description :+1:

PR #224, opened 2016-04-11
Users can discard draft editions

AndrewVos:
If the guide has never been published then the whole guide is destroyed,
otherwise only the latest drafts are destroyed.

Note that the dialog will pop up twice, because of https://github.com/alphagov/service-manual-publisher/pull/225

### DIFF
--- a/app/models/guide.rb
+++ b/app/models/guide.rb
@@ -55,6 +55,7 @@ class Guide < ActiveRecord::Base
 
   def editions_since_last_published
     latest_published_edition = editions.published.last
+    return [] unless latest_published_edition.present?
     editions
       .where("created_at > ?", latest_published_edition.created_at)
   end
--- a/app/models/guide.rb
+++ b/app/models/guide.rb
@@ -53,6 +53,12 @@ class Guide < ActiveRecord::Base
     editions.where(state: "published").any?
   end
 
+  def editions_since_last_published
+    latest_published_edition = editions.published.last
+    editions
+      .where("created_at > ?", latest_published_edition.created_at)
+  end
+
   def work_in_progress_edition?
     latest_edition.try(:published?) == false
   end
--- a/app/models/guide.rb
+++ b/app/models/guide.rb
@@ -4,7 +4,7 @@ class Guide < ActiveRecord::Base
   validate :slug_cant_be_changed_if_an_edition_has_been_published
   validate :latest_edition_has_content_owner, if: :requires_content_owner?
 
-  has_many :editions
+  has_many :editions, dependent: :destroy
   has_one :latest_edition, -> { order(created_at: :desc) }, class_name: "Edition", inverse_of: :guide
 
   accepts_nested_attributes_for :latest_edition
--- a/app/models/publisher.rb
+++ b/app/models/publisher.rb
@@ -24,11 +24,9 @@ class Publisher
       ActiveRecord::Base.transaction do
         publishing_api.discard_draft(content_model.content_id)
 
-        latest_published_edition = content_model.editions.published.last
-        if latest_published_edition.present?
+        if content_model.has_published_edition?
           content_model
-            .editions
-            .where("created_at > ?", latest_published_edition.created_at)
+            .editions_since_last_published
             .destroy_all
         else
           content_model.destroy!
--- a/app/models/publisher.rb
+++ b/app/models/publisher.rb
@@ -19,6 +19,9 @@ class Publisher
     end
   end
 
+  def discard_draft
+  end
+
 private
 
   def save_catching_gds_api_errors(&block)
--- a/app/models/publisher.rb
+++ b/app/models/publisher.rb
@@ -55,9 +55,9 @@ private
   end
 
   class Response
-    def initialize(opts = {})
-      @success = opts.fetch(:success)
-      @error = opts.fetch(:error, nil)
+    def initialize(options = {})
+      @success = options.fetch(:success)
+      @error = options[:error]
     end
 
     def success?
--- a/app/models/publisher.rb
+++ b/app/models/publisher.rb
@@ -34,9 +34,9 @@ class Publisher
           content_model.destroy!
         end
       end
-      DiscardDraftResponse.new(success: true)
+      Response.new(success: true)
     rescue GdsApi::HTTPErrorResponse => e
-      DiscardDraftResponse.new(success: false, errors: e.error_details['error']['message'])
+      Response.new(success: false, error: e.error_details['error']['message'])
     end
   end
 
@@ -48,39 +48,28 @@ private
         if content_model.save
           block.call
 
-          PublicationResponse.new(success: true)
+          Response.new(success: true)
         else
-          PublicationResponse.new(success: false)
+          Response.new(success: false)
         end
       end
     rescue GdsApi::HTTPErrorResponse => e
-      PublicationResponse.new(success: false, errors: e.error_details['error']['message'])
+      Response.new(success: false, error: e.error_details['error']['message'])
     end
   end
 
-  class DiscardDraftResponse
+  class Response
     def initialize(opts = {})
       @success = opts.fetch(:success)
-      @errors = opts.fetch(:errors, [])
+      @error = opts.fetch(:error, nil)
     end
 
     def success?
       @success
     end
 
-    def errors
-      @errors
-    end
-  end
-
-  class PublicationResponse
-    attr_reader :success, :errors
-
-    alias_method :success?, :success
-
-    def initialize(opts = {})
-      @success = opts.fetch(:success)
-      @errors = opts.fetch(:errors, [])
+    def error
+      @error
     end
   end
 end
--- a/app/models/publisher.rb
+++ b/app/models/publisher.rb
@@ -20,22 +20,20 @@ class Publisher
   end
 
   def discard_draft
-    begin
-      ActiveRecord::Base.transaction do
-        publishing_api.discard_draft(content_model.content_id)
+    ActiveRecord::Base.transaction do
+      publishing_api.discard_draft(content_model.content_id)
 
-        if content_model.has_published_edition?
-          content_model
-            .editions_since_last_published
-            .destroy_all
-        else
-          content_model.destroy!
-        end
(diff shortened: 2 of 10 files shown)


app/models/publisher.rb
```diff
@@ -37,6 +58,21 @@ def save_catching_gds_api_errors(&block)
...
+    end
+
+    def success?
+      @success
+    end
+
+    def errors
+      @errors
+    end
+  end
```
thehenster: We could rename `PublicationResponse` to be applicable to both. The interface for the two are nearly identical.

app/models/publisher.rb
```diff
@@ -19,6 +19,27 @@ def publish
...
+            .where("created_at > ?", latest_published_edition.created_at)
+            .destroy_all
+        else
+          content_model.destroy!
+        end
+      end
+      DiscardDraftResponse.new(success: true)
+    rescue GdsApi::HTTPErrorResponse => e
+      DiscardDraftResponse.new(success: false, errors: e.error_details['error']['message'])
+    end
```
thehenster: If we rename `PublicationResponse` with a name you'd feel comfortable using here then this exception rescuing code is shared and could be reused.

app/models/publisher.rb
```diff
@@ -19,6 +19,27 @@ def publish
...
+    begin
+      ActiveRecord::Base.transaction do
+        publishing_api.discard_draft(content_model.content_id)
+
+        latest_published_edition = content_model.editions.published.last
+        if latest_published_edition.present?
+          content_model
+            .editions
+            .where("created_at > ?", latest_published_edition.created_at)
+            .destroy_all
```
thehenster: It'd be nice to remove as much knowledge about the rest of the system as possible here. Making `content_model.lastest_published_edition` (or even `content_model.previously_published?`) and `content_model.editions_since_last_published` kind of thing.

app/models/publisher.rb
```diff
@@ -19,6 +19,27 @@ def publish
...
+        else
+          content_model.destroy!
```
thehenster: Here's a daunting thought.. What if it's in a topic? :(

spec/features/discard_guide_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+    end
+  end
+
+  context "with an unsuccessful discard_draft" do
+    it "does not discard the draft" do
+      guide = create(:guide)
+      discard_draft_response = Publisher::DiscardDraftResponse.new(
+        success: false,
+        errors: "An error occurred",
```
thehenster: Is `errors:` a string? It sounds like it's supposed to be an array. I don't think what I've done for the #save_draft and #publish methods is right. We have to look into that custom exception at some point..

spec/models/publisher_spec.rb
```diff
@@ -87,3 +87,104 @@
...
+    end
+
+    context "when the publishing api call fails" do
+      it "does not destroy anything" do
+        subject = Publisher.new(content_model: guide, publishing_api: publishing_api_which_always_fails)
+        subject.discard_draft
+
+        ids = guide.editions.map(&:id)
+        expect(Guide.where(id: guide.id).count).to eq 1
+        expect(Edition.where(id: ids).count).to eq 4
```
thehenster: You could do `guide.editions.count` or `Edition.where(guide_id: guide.id).count` here to avoid the mapping of the ids.

spec/models/publisher_spec.rb
```diff
@@ -87,3 +87,104 @@
...
+    api
+  end
+
+  let :subject do
+    Publisher.new(content_model: guide, publishing_api: publishing_api)
+  end
+
+  before do
+    allow(publishing_api).to receive(:discard_draft)
+  end
```
thehenster: Could we move the stubbing of this method to the let definition? It just means less jumping around the file for a reader when they drop in to read a single test.

And just to voice a personal preference.. I'm not super pumped for subject blocks either. I constantly forget what the subject is. Is it a publishing api that always passes or a publishing api that always fails? My preference in this case would be a..

```
let :publishing_api_which_always_succeeds do
  publishing_api = double(:publishing_api)
  allow(publishing_api).to receive(:discard_draft)
  publishing_api
end
```
and then just used in the example..
```
it do
  publisher = Publisher.new(content_model: some_guide, publishing_api: publishing_api_that_always_succeeds)
  expect(publisher.discard_draft).to be_success
end
```
You end up with a bit of repetition but I think it's more readable for future generations.

thehenster:
The "Stub out publisher so we don't use the network in tests" commit message makes me think we must have removed Webmock/Fakeweb when we reverted the link checker. We should add it back. It's quite a nice safety net when working on a SOA app like this one.

app/models/publisher.rb
thehenster: I reckon we just validate against it rather than remove it from the topic. Because the relation is in a blob of JSON we can't use the `dependent: :destroy` niceness.

spec/features/discard_guide_spec.rb
AndrewVos: It seems to be a string yes.

thehenster: Your default value for it is an array https://github.com/alphagov/service-manual-publisher/pull/224/files/592d47cbebb8cdf57e7e69192cd8cf1dce8d8abb#diff-33a5f8a8ac794ca434f0c0f60a7d92a3R64

app/models/publisher.rb
AndrewVos: Fair.

app/models/publisher.rb
AndrewVos: Sure, but we don't actually need to do this yet. When we do, we can refactor.

app/models/publisher.rb
AndrewVos: The chances of this happening are fairly low. It would only matter if this was a guide that had never been published.

I think we should leave this for now, refactor Topic to use an actual table, and then revisit it at that point.

AndrewVos: (In the interest of not making this PR too big).

spec/models/publisher_spec.rb
AndrewVos: I chose this approach because it's consistent with the code above. No need to do one thing in two different ways so close together.

AndrewVos: *below

app/models/publisher.rb
AndrewVos: https://trello.com/c/kJUggw1f/315-refactor-topic-tree-into-sql

app/models/publisher.rb
thehenster: I don't think I understand. I'm talking about refactoring the existing functionality. The SQL should be in the model.

app/models/publisher.rb
thehenster: It's possible to add a guide to a topic that hasn't been published and I believe the content team are doing it a lot as they prepare sections to publish all at once. I think the chances of it happening are fairly decent which would leave the data in a broken state.

app/models/publisher.rb
AndrewVos: Cool done.

spec/models/publisher_spec.rb
thehenster: Could we change them all then? It adds an extra query and a little bit more to read. Lots of little improvements make a big improvement in the end etc.

app/models/publisher.rb
```diff
@@ -48,39 +48,28 @@ def save_catching_gds_api_errors(&block)
...
+      Response.new(success: false, error: e.error_details['error']['message'])
     end
   end
 
-  class DiscardDraftResponse
+  class Response
     def initialize(opts = {})
       @success = opts.fetch(:success)
-      @errors = opts.fetch(:errors, [])
+      @error = opts.fetch(:error, nil)
```
thehenster: Having an array allows us to return multiple errors without managing the rendering of them in this object. Say we want to render the errors in `<divs>` with formatting. If returning a string then we have to start joining the errors together here and inserting CSS classes. With an array we can leave that stuff in the view bit.

Empty arrays are nicer to deal with than nils too. Nils can be a pain.

app/models/guide.rb
```diff
@@ -53,6 +53,12 @@ def has_published_edition?
     editions.where(state: "published").any?
   end
 
+  def editions_since_last_published
+    latest_published_edition = editions.published.last
+    editions
+      .where("created_at > ?", latest_published_edition.created_at)
+  end
```
thehenster: This will bork if `latest_published_edition` is nil.

spec/models/publisher_spec.rb
AndrewVos: Done

app/models/publisher.rb
AndrewVos: There's no need for an array here though? There's always only one error returned.

app/models/guide.rb
AndrewVos: Yep, that code path should never be hit.

app/models/publisher.rb
thehenster: But it's this PR that is changing it away from being an array. I'd totally agree if it was already a string/nil. This is changing it for no reason.

Neither is wrong and I've said my piece. If you're very keen to keep this I'm OK with it.

app/models/guide.rb
thehenster: I think that would make sense if we were guaranteed to always have editions in any context a guide was in. We're only guaranteed to not have editions when used in code path in question.

It reduces the reusability of the method.

thehenster:
Thanks for the making the changes with this one :+1:

I think you're refactoring the Topic tree into SQL (https://trello.com/c/kJUggw1f) to handle the removing of guides from topics automatically right? I reckon what you're doing is the best outcome. But if you're keen to get this merged to avoid it failing behind then you could validate against deleting guides if they're in a topic as a temporary measure and then remove the validation later? Up to you. Depends how the refactor is going I guess.

AndrewVos:
Ok I've opened a PR for the topic tree stuff. I would like to merge this for now, rebase that other PR afterwards. Cool?

PR #225, opened 2016-04-11
Don't include jquery twice

AndrewVos:
This was causing data-confirm to fire off two dialogs.

Jquery is already included by `govuk_admin_template`.

Also, we need jquery in jasmine, so import all govuk_admin_template javascript.


thehenster:
LGTM

PR #227, opened 2016-04-12
Store the correct author for guides

AndrewVos:
If a new guide is created then the author is current_user
If a new draft of an already published guide is created then the new
author becomes current_user.

image

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -25,7 +25,11 @@ module ApplicationHelper
     user.name
   end
 
-  def author_options
+  def user_options
     User.pluck(:name, :id)
+  end
+
+  def author_options
+    User.authors.pluck(:name, :id)
   end
 end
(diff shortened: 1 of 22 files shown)


app/helpers/application_helper.rb
```diff
@@ -24,4 +24,8 @@ def css_class_if_current(path, class_name)
   def name_for_user(user)
     user.name
   end
+
+  def author_options
+    User.pluck(:name, :id)
+  end
```
thehenster: 👍

app/helpers/guide_helper.rb
```diff
@@ -15,7 +15,7 @@ def state_label(guide)
   end
 
   def latest_editor_name(guide)
-    guide.latest_edition.user.try(:name).to_s
+    guide.latest_edition.author.try(:name).to_s
```
thehenster: Does it make sense to rename this method too? `#latest_author_name`?

thehenster:
Only 1 minor suggestion. LGTM

app/helpers/guide_helper.rb
AndrewVos: https://github.com/alphagov/service-manual-publisher/pull/227/commits/9d2e94dd8f47c96a6442b251537b0804afa5437b

PR #230, opened 2016-04-13
Rename comments to comments_and_history

AndrewVos:


thehenster:
I think I might rename that route in a separate branch? Theres probably going to be a fair bit of rebasing with our current PRs so I'd prefer to rename after the dust has settled? All good?

AndrewVos:
Yup.. closing

PR #231, opened 2016-04-14
Expand previous editions on history tab

thehenster:
So we can browse the history of a guide we need to be able to open up the previous editions. Previously only the latest edition panel was accessible.

This is part of https://trello.com/c/h4JkfGfP but got missed off.

### DIFF
--- a/spec/features/guide_history_spec.rb
+++ b/spec/features/guide_history_spec.rb
@@ -59,3 +59,23 @@ RSpec.describe "Guide history", type: :feature do
       .reverse
   end
 end
+
+
+RSpec.describe "Guide history", type: :feature do
+  scenario "viewing previous editions" do
+    guide = create(:published_guide)
+    guide.editions << build(:edition, version: 2)
+
+    visit edition_comments_path(guide.reload.latest_edition)
+
+    expect_edition_to_be_open("Version #2")
+
+    click_link "Version #1"
+
+    expect_edition_to_be_open("Version #1")
+  end
+
+  def expect_edition_to_be_open(title)
+    expect(page).to have_css(".open-edition .panel-heading", text: title)
+  end
+end
--- a/spec/features/guide_history_spec.rb
+++ b/spec/features/guide_history_spec.rb
@@ -7,19 +7,23 @@ RSpec.describe "Guide history", type: :feature do
     stub_publisher
     create_guide_community
 
-    travel_to "2004-11-24 01:04:44".to_time do
+    travel_to "2004-11-24".to_time do
       visit root_path
       click_on "Create a Guide"
       fill_out_new_guide_fields
       click_first_button "Save"
+    end
 
+    travel_to "2004-11-25".to_time do
       click_on "Comments and history"
 
       within ".open-edition" do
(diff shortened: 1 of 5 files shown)


spec/features/guide_history_spec.rb
```diff
@@ -59,3 +63,23 @@ def events
...
+
+    expect_edition_to_be_open("Version #2")
+
+    click_link "Version #1"
+
+    expect_edition_to_be_open("Version #1")
+  end
+
+  def expect_edition_to_be_open(title)
+    expect(page).to have_css(".open-edition .panel-heading", text: title)
```
AndrewVos: Is this `open-edition` class only used for tests?

AndrewVos: As it is, this isn't a very good feature test. Rather test that something is appearing on the page?

AndrewVos: Oh I didn't see the title part. I think we just need that assertion

thehenster: Sorry - I don't think I understand about just needing that assertion. Could you rephrase me?

thehenster: Yeah, the CSS class `open-edition` makes it easier to test. Personally I'm comfortable with it. It's a pretty semantic block name and isn't likely to change much and so will be less likely to break the test than, for example, we check which type of chevron is used to determine open-ness. We make little compromises all over to increase the testability of code.

I'm open to suggestions - what would you prefer?

thehenster: rephrase *for me? .. Soz. I miss out words sometimes.

AndrewVos: I meant that just testing for the title is enough in this test. Just realised that the title is displayed for all versions though so that's not really enough.

In a feature test you really want to be testing what actually happens to the page. Testing that something has a class isn't really enough. The test should be asserting that the content of that expanded section is actually visible.
If we deleted all the code inside that panel, in the view, this test would be green.

AndrewVos:
Other than one comment, this lgtm.

AndrewVos:
Didn't you want to rename this view/url as part of this pull request?

spec/features/guide_history_spec.rb
thehenster: The selector is checking the title but it's scoped to be nested under `.open-edition`.. so I suppose my assertion is saying.. "If this title heading can be found in an `open-edition` block then consider it to be open". But they could all be open right? My test doesn't say that the previous one is closed. I'll add that - thanks.

Do you feel strongly about not using the `open-edition` to indicate open-ness? Like I was saying above, I find it to be a good compromise.. Perhaps an alternative would be to check if the bootstrap class `.panel-body` is present? It looks like collapsed ones don't have one. Makes sense to me.. A collapsed thing doesn't have a body. Another one could be that it contains some nested `.event` divs? A collapsed thing shouldn't contain any events. What do you think?

thehenster:
:wave: @AndrewVos. I've added a commit for the rework of that test.

AndrewVos:
Nice! Lgtm

PR #232, opened 2016-04-15
Store topics in SQL instead of a JSON column

AndrewVos:


app/models/guide.rb
```diff
@@ -37,12 +37,14 @@ def latest_persisted_edition
...
-                 end.any?
-               end.first
+    @topic ||= (
+      topic_section_guide = TopicSectionGuide.where(guide: self).first
+      if topic_section_guide.present?
+        topic_section_guide.topic_section.topic
+      else
+        nil
+      end
+    )
```
thehenster: How about using a `JOIN` here?

AndrewVos: https://github.com/alphagov/service-manual-publisher/pull/232/commits/7ffde9e7765f8064f55ce923700028c6ba8b9c0a

db/migrate/20160413135658_create_topic_section.rb
```diff
@@ -0,0 +1,11 @@
+class CreateTopicSection < ActiveRecord::Migration
+  def change
+    create_table :topic_sections do |t|
+      t.integer :topic_id, null: false
+      t.string :title, null: false
+      t.string :description, null: false
```
thehenster: I 10% remember that Luke asked if he could create sections without giving a title and description - and he could. I have this feeling we might have some sections without headings..

AndrewVos: Good catch! We do!

AndrewVos: OK allowing nulls.

app/models/topic.rb
```diff
@@ -22,14 +24,9 @@ def slug
...
-      grouping['guides'].map do |guide_id|
-        Integer(guide_id)
-      end
-    end.flatten.uniq
-    Guide
-      .where(id: ids)
-      .pluck(:content_id)
+    topic_sections.map do |topic_section|
+      topic_section.guides.map(&:content_id)
+    end.flatten
```
thehenster: If we defined a `has_many :guides, through: :topic_sections` then this method could become `guides.map(&:content_id)` which would be nice.

It looks like the `uniq` is gone - do we no longer need it here?

app/presenters/topic_presenter.rb
```diff
@@ -64,13 +60,8 @@ def content_owner_content_ids
...
-      ids = topic_groups.flat_map { |group| group['guides'] }.uniq
-      Guide.where(id: ids)
-    end
-  end
-
-  def topic_groups
-    @topic_groups ||= topic.tree.select { |h| h.is_a?(Hash) }
+    topic.topic_sections.map do |topic_section|
+      topic_section.guides.to_a
+    end.flatten.uniq
```
thehenster: The extra has_many through would shorten this bit too I think.

db/migrate/20160413140153_create_topic_section_guides.rb
```diff
@@ -0,0 +1,10 @@
+class CreateTopicSectionGuides < ActiveRecord::Migration
+  def change
+    create_table :topic_section_guides do |t|
+      t.integer :topic_section_id, null: false
+      t.integer :guide_id, null: false
```
thehenster: Indexing these two columns would be good. They'll be used a lot because of the has_many :through.

app/controllers/topics_controller.rb
```diff
@@ -75,6 +85,21 @@ def update_topic_params
   end
 
   def updatable_topic_attributes
-    [:title, :description, :tree, content_owner_ids: []]
+    [:title, :description, content_owner_ids: []]
+  end
+
+  def store_topic_sections_in_topic
+    @topic.topic_sections.destroy_all
```
thehenster: Will the related `TopicSectionGuide`s get blown away with this? I can't see the dependent: :destroy but I might be missing something.

thehenster:
I like the approach :+1: Trying to reshuffle everything on a change would be much more complicated than blowing everything away and rebuilding.

app/controllers/topics_controller.rb
AndrewVos: https://github.com/alphagov/service-manual-publisher/pull/232/commits/acfb0d9fde2795520c9cbcfcf77c682f8eaced51

Look good?

AndrewVos:
@thehenster rereview pls

db/migrate/20160413140153_create_topic_section_guides.rb
AndrewVos: Done in seperate migration

thehenster:
Merge when you're ready :+1:

PR #236, opened 2016-04-20
Show more information about guide editions in "comments and history"

AndrewVos:
Very simple CSS for now:

I should probably extract a partial, but I would prefer to rename the whole comments thing to comments_and_history or something like that first.

image


thehenster:
Defo needs a partial but if you've got a plan for it in a later PR then merge away :+1:

PR #238, opened 2016-04-21
Points page

thehenster:
This only covers the points pages and not the service standard they will eventually influence. https://trello.com/c/YJnXkLQ1

### DIFF
--- a/spec/controllers/guides_controller_spec.rb
+++ b/spec/controllers/guides_controller_spec.rb
@@ -67,4 +67,12 @@ RSpec.describe GuidesController, type: :controller do
       end
     end
   end
+
+  describe "a malicious user trying to initialise an unwanted Guide STI constant" do
+    it "defaults to a Guide" do
+      get :new, type: 'Module'
+
+      expect(assigns[:guide_form].guide.class).to eq(Guide)
+    end
+  end
 end
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -17,7 +17,7 @@ class GuidesController < ApplicationController
   end
 
   def new
-    type = params[:community].present? ? 'GuideCommunity' : nil
+    type = params[:type].presence_in(%w{ GuideCommunity Point })
 
     @guide_form = GuideForm.new(
       guide: Guide.new(type: type),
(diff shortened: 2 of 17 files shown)


thehenster:
@AndrewVos What do you think about the column name `editions.summary`? It represents the bit of text under the title here http://service-manual-staging.herokuapp.com/service-standard/point-1/

We could call it `editions.lede` or `editions.standfirst` but those terms are more about layout rather than what the information is.. I don't have any good ideas.

AndrewVos:
I think summary is good for now. Column names can be changed really easily. Merge it !

AndrewVos:
Though, it's kind of weird calling the type "Point" isn't it. Would be nice if it was PointPage, but then we'll have to go around suffixing all types with Page.

spec/controllers/guides_controller_spec.rb
```diff
@@ -67,4 +67,12 @@
       end
     end
   end
+
+  describe "a malicious user trying to initialise an unwanted Guide STI constant" do
```
AndrewVos: Do we really need this test? I think it's safe enough because we construct the type in code in the controller #new?

AndrewVos:
lgtm. Some comments above but nothing to stop this going through.

thehenster:
Thanks @AndrewVos. Yeah, I agonised over the Point/Point Page thing too and reached the same conclusion as you :) "Create a Guide Page" and "Create a Guide Community Page" would be worse.

spec/controllers/guides_controller_spec.rb
thehenster: I added this test to drive adding `params[:type].presence_in(%w{ GuideCommunity Point })` which is defending against someone trying to slip something into `Guide.new(type: type)` which, deep deep down, is essentially performing an eval on params 😱 

It's probably too paranoid - I kind of the like the paranoia here though. Evalling params scares me!

AndrewVos: There's an eval on params??

thehenster: Yeah - somewhere that `params[:type]` string will need to be converted into a constant for the STI to do it's thing.

```
irb(main):003:0> Guide.new(type: 'Point')
=> #<Point id: nil, slug: nil, content_id: nil, created_at: nil, updated_at: nil, tsv: nil, type: "Point">
```

Rails probably has some kind of protection around it.. but I just don't know what it is. It'll error if you try and initialise a class that isn't a subclass of Guide. But what else could you try and throw in there? Scary.

AndrewVos: Agree.

AndrewVos:
Merge it!

PR #239, opened 2016-04-22
User can unpublish guides that have been published previously

AndrewVos:
A guide can be unpublished, which means redirecting that guide to either a
topic or a guide.

- New Edition state "unpublished"
- Edition action buttons disabled when a guide is unpublished
- Edition inputs are disabled when a guide is unpublished
- Store redirect in the publishing api
- Store new author as the person who unpublished
- Rename SlugMigrationPublisher to RedirectPublisher

image

image

image

image

## Unpublish button is now floated right:

image

## Unpublished state is gray:

image

## Edition actions are now reordered from:

```
Preview
Save
Discard new guide
Discard draft
Unpublish
Send for review
Approve for publication
Publish
Send for review
Back to all guides
```

## to:

```
Preview
Save
Unpublish
Send for review
Approve for publication
Publish
Send for review
Discard new guide
Discard draft
Back to all guides
```

image


app/models/slug_migration_publisher.rb
```diff
@@ -1,24 +1,24 @@
 require "gds_api/publishing_api_v2"
 
 class SlugMigrationPublisher
-  def process(slug_migration)
+  def process(content_id:, old_path:, new_path:)
```
thehenster: I reckon it'd be nice to rename this to `RedirectPublisher` as it's no longer about SlugMigrations?

It'd be good to use the `PUBLISHING_API` constant too. Maybe it could be a constructor argument like in `Publisher`?

AndrewVos: You're looking at old code ;)

app/controllers/slug_migrations_controller.rb
```diff
@@ -39,7 +39,7 @@ def update
 
     ActiveRecord::Base.transaction do
       if @slug_migration.update_attributes(slug_migration_parameters)
-        SlugMigrationPublisher.new.process(
+        RedirectPublisher.new.process(
```
thehenster: :)

app/models/guide.rb
```diff
@@ -50,6 +50,15 @@ def has_published_edition?
     editions.where(state: "published").any?
   end
 
+  def has_unpublished_edition?
+    editions.where(state: "unpublished").any?
+  end
```
thehenster: This method could be private?

app/models/guide.rb
```diff
@@ -50,6 +50,15 @@ def has_published_edition?
...
+
+  def can_be_unpublished?
+    return false if has_unpublished_edition?
+    return true if has_published_edition?
+  end
```
thehenster: A totally subjective one..```def can_be_unpublished?
  has_published_edition? && !has_unpublished_edition?
end```?

AndrewVos: Mine is prettier!

AndrewVos: Maybe something like `has_been_published_before? && has_never_been_unpublished?`

thehenster: Yep :+1:

app/views/editions/_actions.html.erb
```diff
@@ -5,37 +5,52 @@
       <% if edition.persisted? %>
         <%= link_to "Preview", preview_content_model_url(guide), class: 'btn btn-default add-right-margin' %>
       <% end %>
+
+      <% if !publish_controls_only %>
+        <%= form.submit "Save", class: 'btn btn-default add-right-margin js-ok-to-navigate-away', name: :save, disabled: edition.unpublished?, data: disable_if_new_record(guide) %>
```
thehenster: Could the disabled and data-disabled be combined?

Actually - I don't think the data-disable is doing anything. On integration the button seems to be available when the record is new.

AndrewVos: Yeah I don't think that does what whoever wrote it wanted

AndrewVos: Deleted it.

app/views/unpublishes/new.html.erb
```diff
@@ -0,0 +1,22 @@
...
+
+    <div class="form-group">
+      <%= f.label :new_path do %>
+        Redirect to (<span class="text-info">only published guides and topics will appear in this list</span>)
+      <% end %>
+      <%=
+        f.select :new_path,
+          grouped_options_for_select(@select_options),
+          {include_blank: "Please choose a path to redirect to"},
+          {class: "select2 slug-migration-redirect-to input-md-12 form-control"}
```
thehenster: I can't see what the `slug-migration-redirect-to` class does here or in the original slug migrations actually 😕  A left over from the past?

AndrewVos: Yep, killing it

app/models/guide.rb
AndrewVos: Nah, cause the one above it isn't. I would rather be consistent.

thehenster: Oh.. could they both be moved to private then? If we're using the methods anywhere outside the class they should be private. Right?

AndrewVos: The other method is used outside the class, which is why I kept them both
right next to each other, more consistent.

thehenster:
Go for it :+1:

PR #249, opened 2016-05-03
Republish all guides

thehenster:
We need to be able to republish all of our content. We will soon be changing frontend which requires us to republish all content to change the rendering_app.

It is not possible to republish old editions from the UI and we don't want to make our content designers finish all of their work at the same time to facilitate clicking the publish button. We therefore need a script that can republish the correct live edition.

### DIFF
--- a/lib/tasks/republish.rake
+++ b/lib/tasks/republish.rake
@@ -1,0 +1,8 @@
+namespace :republish do
+  desc "republish all guides"
+  task guides: :environment do
+    Guide.find_each do |guide|
+      GuideRepublisher.new(guide).republish
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/tasks/republish.rake
```diff
@@ -0,0 +1,8 @@
+namespace :republish do
+  desc "republish all guides"
+  task guides: :environment do
+    Guide.find_each do |guide|
+      GuideRepublisher.new(guide).republish
```
AndrewVos: Maybe some output here because the publisher is really slow. Going to merge it anyway though... nice to add in the future though.

AndrewVos:
Nice small and sweet PR.

lib/tasks/republish.rake
thehenster: Good idea: https://github.com/alphagov/service-manual-publisher/pull/252

PR #253, opened 2016-05-05
Generate title+full slug automatically from Topic Section and Title

AndrewVos:


thehenster:
Yo @AndrewVos. I've added the changes I wanted for the GuideForm. I've swapped around the title slug and topic section fields and I've sent an email to Mike (and CC'd you) about adding the `->` to the option text in the dropdown too.

But.. I think we can merge this now if you're happy?

app/assets/javascripts/slug-generation.js
```diff
@@ -1,17 +1,48 @@
 $(function() {
+  var $form = $(".js-guide-form");
+
+  if ($form.length == 0) {
```
thehenster: I don't think you need the `== 0` comparison.. `0 == false` in JS.

app/assets/javascripts/slug-generation.js
```diff
@@ -1,17 +1,48 @@
...
+    if (!topicPath) { return }
+    slug += topicPath + "/";
+
+    if (!$titleSlug.val()) { return }
+    slug += slugify($titleSlug.val())
+
+    $slug.val(slug);
+  }
+
+  generateSlug();
```
thehenster: I'd find it easier to follow the flow if this `generateSlug()` call was with the rest of the stuff above.

app/assets/javascripts/slug-generation.js
thehenster: It'd be nice if this was a function that explained what it was doing.. Like..

```
if (!formPresent) return;
```

app/assets/javascripts/slug-generation.js
```diff
@@ -1,17 +1,48 @@
...
+
+    var topicPath = $topicSection.find(":selected").parent().data("path");
+    if (!topicPath) { return }
+    slug += topicPath + "/";
+
+    if (!$titleSlug.val()) { return }
+    slug += slugify($titleSlug.val())
+
+    $slug.val(slug);
+  }
```
thehenster: This confu3es me.. Sorry. Could it be reorganised or commented somehow? I don't have any bright ideas..

app/assets/javascripts/slug-generation.js
AndrewVos: How about just removing it entirely. It doesn't really add much in terms of performance, and the code below will not break without it: https://github.com/alphagov/service-manual-publisher/pull/253/commits/ac41101eaa6883f483878373f4437b0c7c8823c9

AndrewVos: Oh whoops, wrong file!

app/assets/javascripts/slug-generation.js
AndrewVos: It can't be moved up there because `generateSlug()` is defined below it, and wouldn't be available at this point.

thehenster:
Merge when you're ready :+1:

PR #256, opened 2016-05-09
Fix guide index not filtering by state correctly.

AndrewVos:
We were filtering by guides that ever had an edition of the requested
state, not guides with a latest edition that is in the state.

### DIFF
--- a/app/models/guide.rb
+++ b/app/models/guide.rb
@@ -7,8 +7,16 @@ class Guide < ActiveRecord::Base
 
   has_many :editions, dependent: :destroy
 
+  scope :only_latest_edition, -> {
+    joins(:editions)
+      .where('editions.created_at = (SELECT MAX(editions.created_at) FROM editions WHERE editions.guide_id = guides.id)')
+  }
+
+  scope :in_state, -> (state) {
+    only_latest_edition.where("editions.state = '#{state}'") if state.present?
+  }
+
   scope :by_author, ->(author_id) { where(editions: { author_id: author_id }) if author_id.present? }
-  scope :in_state, ->(state) { where(editions: { state: state }) if state.present? }
   scope :owned_by, ->(content_owner_id) { where(editions: { content_owner_id: content_owner_id }) if content_owner_id.present? }
 
   delegate :title, to: :latest_edition
--- a/app/models/guide.rb
+++ b/app/models/guide.rb
@@ -12,12 +12,15 @@ class Guide < ActiveRecord::Base
       .where('editions.created_at = (SELECT MAX(editions.created_at) FROM editions WHERE editions.guide_id = guides.id)')
   }
 
-  scope :in_state, -> (state) {
+  scope :in_state, ->(state) {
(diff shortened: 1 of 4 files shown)


app/models/guide.rb
```diff
@@ -7,9 +7,20 @@ class Guide < ActiveRecord::Base
 
   has_many :editions, dependent: :destroy
 
-  scope :by_author, ->(author_id) { where(editions: { author_id: author_id }) if author_id.present? }
-  scope :in_state, ->(state) { where(editions: { state: state }) if state.present? }
-  scope :owned_by, ->(content_owner_id) { where(editions: { content_owner_id: content_owner_id }) if content_owner_id.present? }
+  scope :only_latest_edition, -> {
+    joins(:editions)
+      .where('editions.created_at = (SELECT MAX(editions.created_at) FROM editions WHERE editions.guide_id = guides.id)')
```
thehenster: I loaded it in a console to eyeball it.. Seems fine to me.

```
SELECT "guides".*
FROM "guides"
INNER JOIN "editions"
ON "editions"."guide_id" = "guides"."id"
WHERE (
  editions.created_at = (
    SELECT MAX(editions.created_at)
    FROM editions
    WHERE editions.guide_id = guides.id
  )
)
AND (editions.state = 'published')
```

app/models/guide.rb
```diff
@@ -7,9 +7,20 @@ class Guide < ActiveRecord::Base
 
   has_many :editions, dependent: :destroy
 
-  scope :by_author, ->(author_id) { where(editions: { author_id: author_id }) if author_id.present? }
-  scope :in_state, ->(state) { where(editions: { state: state }) if state.present? }
```
thehenster: Looks like the original would have errored unless the join was added elsewhere..

spec/features/filter_spec.rb
```diff
@@ -2,23 +2,44 @@
...
+        build(:edition, state:"unpublished", title: "Unpublished"),
+      ],
     )
 
     filter_by_state "Draft"
-    expect(page).to_not have_text "Edition 1"
-    expect(page).to have_text "Edition 2"
+    expect(all_titles).to eq [
+      "Draft"
+    ]
```
thehenster: I find this test could be more explicit.. There are three potential draft editions created and we're just asserting any of them appear in the list.

I'd be more explicit if the titles were set when creating the factories and we checked that the specific title was returned in the list?

AndrewVos: I agree...

spec/models/guide_spec.rb
```diff
@@ -288,3 +288,90 @@
     expect(guide.live_edition).to eq(nil)
   end
 end
+
+RSpec.describe Guide, "#in_state" do
```
thehenster: Scopes are (eventually) class methods so should read `".in_state"`. Same for the describe blocks below.

thehenster:
LGTM

PR #257, opened 2016-05-10
Add missing indexes

AndrewVos:


thehenster:
Something odd going on here.. Did you branch from master?

I'm up for the indexes if they're genuine. I'm less up for leaving that gem included in the Gemfile. It doesn't seem super mature.

AndrewVos:
This was an oldddddd branch and I was cleaning up some branches. Will take
a look in a bit

PR #260, opened 2016-05-17
Replace rummageable with gds api adapters

thehenster:
Deleting a document from rummager wasn't working with Rummageable because it was impossible to delete a document with a type other than `edition` or `best_bet`.

Rummageable is a deprecated thing and so we switch to using gds-api-adapaters.


AndrewVos:
lgtm

PR #261, opened 2016-05-17
A temporary task to clean up some unpublished guides

thehenster:
We had a bug that didn't delete unpublished guides from search. The bug is fixed for future guides. This cleans up the ones that failed and removes them from the search index.


AndrewVos:
Lgtm!

PR #263, opened 2016-05-18
Sanitize the rendering of an unescaped parameter

thehenster:
Pretty paranoid but just in case there is something potentially dangerous in the user's filename.


AndrewVos:
Doesn't this happen regardless?

PR #265, opened 2016-05-19
Remove the filename from the failed upload alert

thehenster:
There was a small risk of rendering unsanitized input and having the filename in the alert wasn't a huge benefit as the user will know which file they have just tried to upload.


AndrewVos:
:heart:

PR #271, opened 2016-05-23
Allow multi-line comments

AndrewVos:
Store the contents as non-html string, but when rendering preserve new
lines.


thehenster:
:+1:

We need to add the line breaks to the email notification too. Separate PR?

PR #272, opened 2016-05-24
Just delete the duplicated technology topic

thehenster:
This will leave the topic content item in the publishing-api but we have a bunch of these. We'll address removing them at a later date.


AndrewVos:
Kind of feel like it would be safer to delete by content_id. Not really sure why though. I suppose the chances of something having a duplicate ID are zero.

AndrewVos:
lgtm

PR #274, opened 2016-05-24
"View changes" links for past versions should have a valid diff link

AndrewVos:
We were only showing a diff link for the latest version on the history tab.

Here we link to valid diffs for all versions and also remove the branching in the template which probably contributed to this bug slipping through the cracks.


dougdroper:
LGTM 👍

PR #275, opened 2016-05-24
Track who makes changes to guides

thehenster:
https://trello.com/c/POsJcfsx


db/migrate/20160520134625_add_created_by_to_editions.rb
```diff
@@ -0,0 +1,5 @@
+class AddCreatedByToEditions < ActiveRecord::Migration
+  def change
+    add_column :editions, :created_by_id, :integer
```
AndrewVos: `null: false`

AndrewVos:
lgtm, other than one comment

db/migrate/20160520134625_add_created_by_to_editions.rb
dougdroper: index?

thehenster: Thanks @AndrewVos. I'll add that

I don't think it's involved in a query is it @dougdroper? Do you defensively add them?

AndrewVos: Might as well just add it anyway.

PR #277, opened 2016-06-06
Tag all content to GDS in publishing-api

tijmenb:
This app currently [tags all content to the GDS org in Rummager](https://github.com/alphagov/service-manual-publisher/blob/f434aa0aa2671
3e829c93b838031dc0abfd2fe25/app/models/guide_search_indexer.rb#L24).

In the near future, Rummager will start indexing the `links` hash from the publishing-api. This means that we should start sending the `organisations` tagging to publishing-api too, so that rummager can copy this.

When we turn on the links-indexing the `organisations` tagging can be removed from the rummager payload.

https://trello.com/c/2PX73XvD

### DIFF
--- a/app/presenters/guide_presenter.rb
+++ b/app/presenters/guide_presenter.rb
@@ -30,7 +30,9 @@ class GuidePresenter
   end
 
   def links_payload
-    links = {}
+    links = {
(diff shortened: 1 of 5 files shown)


tijmenb:
(To be reviewed by @thehenster)

PR #278, opened 2016-06-07
Set x_govuk_authenticated_user header...

steventux:
https://trello.com/c/6nDCqZXi/729-user-id-into-event-log
The publishing API event log should have a reference to the authenticated user
so ensure that the appropriate GdsApi::GovukHeaders header is set to meet expectations
of the publishing API event logging.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,9 +5,16 @@ class ApplicationController < ActionController::Base
 
   include GDS::SSO::ControllerMethods
   before_filter :require_signin_permission!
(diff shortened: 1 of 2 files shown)


thehenster:
Cool. Does this mean we're going to start storing who did what in the publishing api @steventux?

steventux:
At a very basic level yes, [the `EventLogger` will store the uuid against each event](https://github.com/alphagov/publishing-api/blob/master/app/event_logger.rb#L12),  so more of an internal audit trail for now. This doesn't yet address the publishing app's needs for change history and audit info.

PR #283, opened 2016-06-15
Guide manager

thehenster:
This splits the various actions taken on a guide into two groups.

1) Saving guide content - validating, creating and updating the various fields eg. body, title, description etc.
2) Managing the state of the guide - requesting a review, approving, publishing and discarding

Saving content is handled by the GuideForm. Managing the state is handled by the GuideManager.

This removes more logic from the controller and hopefully the saving vs. managing language makes the guide controller easier to follow.

The previous Publisher object was too optimistic that it was going to be general enough to work for guides and topics. The symptoms of this was that additional functionality started to appear on Publisher that wasn't relevant for topics. eg. Publisher#discard_draft. And when we wanted to extend shared functionality, but only for one type, the controller became a dumping ground. Publisher has now become TopicPublisher as it only deals with Topics.


app/controllers/guides_controller.rb
```diff
@@ -6,196 +6,98 @@ def index
...
     end
   end
 
 private
 
-  def send_for_review(guide, edition)
-    edition.state = 'review_requested'
-    edition.created_by = current_user
-    edition.save!
+  def manage!(guide, action, opts = {})
```
dougdroper: it feels like a lot of duplication between this and manage. Is there a way to reduce it?

app/models/guides_filter.rb
```diff
@@ -0,0 +1,55 @@
...
+    'content_owner',
+    'page',
+    'page_type',
+    'q',
+    'state'
+  ]
+
+  def initialize(scope)
+    @scope = scope
+    # TODO: :content_owner not being included is resulting in an N+1 query
```
dougdroper: What does this mean? Why are you not including it?

thehenster: I just haven't got round to fixing it yet.. content_owner joins `guides` and if the user is searching, using the query in `Guide.search`, then the reference to `tsv` is ambiguous. You end up returning guides because the search matched it's content_owner.

It just needs to be preloaded separately..

app/controllers/guides_controller.rb
thehenster: I wanted to do that to :) I couldn't find a more readable way and got bored because it was only a few lines. Got a suggestion for me? (Don't waste your time on it though..)

app/models/guides_filter.rb
dougdroper: 👍

app/controllers/guides_controller.rb
dougdroper: yeah I was thinking along the lines of 
`manage(guide, "request_review!", etc)` to remove the need to the two
but that is a bit hand wavy, and not such a big deal

dougdroper:
I think it is a clear separation, and a big improvement 👍

PR #284, opened 2016-06-15
Delete the `topics` field from guide links

thehenster:
This should be removed after we're republished.


mobaig:
Merging - discussed with @thehenster IRL, and this is a temporary change that's going to be reverted shortly.

PR #287, opened 2016-06-21
Update readme

36degrees:
- Fix grammar
- Use db:reset instead of create and migrate
- Update instructions to make bowler the preferred method of running publisher


thehenster:
Nice improvement :+1:

PR #288, opened 2016-06-21
Document the guide's nested URL structure

thehenster:
I'd like to get Sean's feedback on the performance analysis section..

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ Service Manual Publisher is in very early stages and is going to be used for pub
 
 PostgreSQL-backed Rails 4 "Publishing 2.0" application for internal use, with no public facing aspect.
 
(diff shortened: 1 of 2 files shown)


thehenster:
Hi @deborahchua and @36degrees,

I _was_ going to add an additional bit to this doc about a type of content that doesn't have a topic which is just emerging now. However, it is just emerging, so I'd rather merge this now and add to it later if it becomes a thing.

👀 welcome on this!

thehenster:
Sean has OK'd the performance section btw.

36degrees:
Mentioned yesterday about the 'workaround' for changing a URL (unpublishing, cloning to a new guide) having undesirable side effects, e.g. loss of edit history. Worth fleshing that out a bit? At the minute it just sounds like it's a faff.

thehenster:
Sure :+1:. Coming up.

thehenster:
I've added an extra point about maintaining additional redirects too..

36degrees:
LGTM 👍

PR #289, opened 2016-06-22
Revert "Delete the `topics` field from guide links"

thehenster:
This was a temporary change required to purge all of the `topics` fields from the links. Since we're no longer creating the `topics` field we no longer have to reset it each time we patch the links.


36degrees:
👍

PR #290, opened 2016-06-23
Present multiline comments correctly in HTML emails

36degrees:
Use simple_format to automatically wrap comment text in paragraph tags / use line breaks in the HTML part of new comment notifications

### DIFF
--- a/spec/mailers/notification_mailer_spec.rb
+++ b/spec/mailers/notification_mailer_spec.rb
@@ -26,6 +26,29 @@ RSpec.describe NotificationMailer, type: :mailer do
         expect(part.body.to_s).to include "guides/#{guide.id}/editions"
       end
     end
+
+    it "presents multi line comments correctly" do
+      comment_text = %{This guide sure could use more cow bell.
+Cow bell makes everything better!
+
+Much better.}
+      formatted_comment = %{<p>This guide sure could use more cow bell.
+<br />Cow bell makes everything better!</p>
+
+<p>Much better.</p>}
+
+      comment = edition.comments.create!(comment: comment_text, user: luke)
+
+      email = NotificationMailer.comment_added(comment).deliver_now
+
+      # The HTML part of the email should include the comment wrapped in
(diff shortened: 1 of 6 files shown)


spec/mailers/notification_mailer_spec.rb
```diff
@@ -26,6 +26,29 @@
         expect(part.body.to_s).to include "guides/#{guide.id}/editions"
       end
     end
+
+    it "presents multi line comments correctly" do
+      comment_text = %{This guide sure could use more cow bell.
```
36degrees: Definitely open to suggests on how to indent this so it's not awful.

36degrees:
Caveats / questions:

- I've not run this in the browser / actually sent myself an email, this is all based on tests so far.
- I believe that `simple_format` is safe to use and isn't going to start unescaping other HTML tags etc, but it'd be good to get someone to double check that.
- Do we want to update the way we present comments on the web view to `simple_format` for consistency?

thehenster:
If you don't want to set up mail delivery on your local then you can use something like https://github.com/ryanb/letter_opener or https://mailcatcher.me/. Or we can deploy a specific branch to integration to test there as there probably won't be too many surprises here.

I'm not sure I understand the worry about `simple_format` unescaping stuff?

I think it'd be nice to use `simple_format` in the web view too.

Well done remembering about the text version of the email! :clap:

spec/mailers/notification_mailer_spec.rb
thehenster: There isn't a nice way to do this really. Some people roll their own thing that strips leading whitespace.. but.. seems a bit OTT for the benefit to me.

Ruby 2.3 has a thing but we're on 2.2. https://infinum.co/the-capsized-eight/articles/multiline-strings-ruby-2-3-0-the-squiggly-heredoc

Maybe we can upgrade? :shrug:

36degrees:
Tested locally and it appears to be working fine.

thehenster:
:+1:

PR #291, opened 2016-06-27
Service standard

thehenster:
We are replacing https://www.gov.uk/service-manual/digital-by-default.

https://trello.com/c/oDSuK5ay

Broadly speaking this PR..

* Publishes a new schema service_manual_service_standard
* Removes the need for a topic for Points

We are storing the relationship between the standard and the points in the details hash of the service_manual_service_standard. This will be moved to the links hash when we can take advantage of custom fields in link expansion.

### DIFF
--- a/app/helpers/guide_helper.rb
+++ b/app/helpers/guide_helper.rb
@@ -36,10 +36,12 @@ module GuideHelper
 
   def topic_section_options_for_select
     Topic.includes(:topic_sections).map do |topic|
+      relative_path = topic.path.gsub("/service-manual", "")
+
       [
         topic.title,
         topic.topic_sections.map { |ts| ["#{topic.title} -> #{ts.title}", ts.id] },
-        { "data-path" => topic.path }
+        { "data-path" => relative_path }
       ]
     end
   end
--- a/app/helpers/guide_helper.rb
+++ b/app/helpers/guide_helper.rb
@@ -36,7 +36,7 @@ module GuideHelper
 
   def topic_section_options_for_select
     Topic.includes(:topic_sections).map do |topic|
-      relative_path = topic.path.gsub("/service-manual", "")
+      relative_path = topic.path.sub("/service-manual", "")
 
       [
         topic.title,
--- a/app/helpers/slug_migrations_helper.rb
+++ b/app/helpers/slug_migrations_helper.rb
@@ -1,0 +1,18 @@
+module SlugMigrationsHelper
+  def select_options
+    guide_select_options = Guide
+      .with_published_editions
+      .order(:slug).pluck(:slug)
+      .map { |g| [g, g] }
+
+    topic_select_options = Topic
+      .order(:path).pluck(:path)
+      .map { |g| [g, g] }
+
+    {
+      "Other" => ["/service-manual"],
+      "Topics" => topic_select_options,
+      "Guides" => guide_select_options,
+    }
+  end
+end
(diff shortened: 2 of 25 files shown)


app/helpers/guide_helper.rb
```diff
@@ -36,10 +36,12 @@ def guide_community_options_for_select
 
   def topic_section_options_for_select
     Topic.includes(:topic_sections).map do |topic|
+      relative_path = topic.path.gsub("/service-manual", "")
```
36degrees: Probably unlikely, but I think if we had a meta topic e.g. 'Service Manual Usage' that lived at '/service-manual/service-manual-usage` we might break things with this. Regex with start of string anchor?

36degrees: Or, just use `sub` rather than `gsub` might be OK if we're sure topic.path will _always_ start with `/service-manual`.

app/helpers/slug_migrations_helper.rb
```diff
@@ -0,0 +1,18 @@
+module SlugMigrationsHelper
+  def select_options
+    guide_select_options = Guide
+      .with_published_editions
+      .order(:slug).pluck(:slug)
+      .map { |g| [g, g] }
```
36degrees: Did we want to revisit this?

spec/presenters/service_standard_presenter_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "rails_helper"
+
+RSpec.describe ServiceStandardPresenter, "#content_id" do
+  it "is a preassigned UUID" do
```
36degrees: mix of `is a` and `returns a` for methods that just 'return a thing', do we have a standard way of describing return values? (Bordering on pedantry? 😛)

app/helpers/guide_helper.rb
thehenster: Thanks - I've gone for the `sub` version as it'll hopefully be easier to use when we go around moving all the "/service-manual" strings we have lurking into config.

app/helpers/slug_migrations_helper.rb
thehenster: I'd prefer to leave it for this PR. I've added a placeholder item below the icebox line to remind us to pick it up.

spec/forms/guide_form_spec.rb
```diff
@@ -404,3 +418,11 @@
     expect(guide_form.to_param).to eq("5")
   end
 end
+
+RSpec.describe GuideForm, "#slug_prefix" do
+  it "is /service-manual" do
```
36degrees: Sorry, probably should've pointed this one out as well (on the returns / is a front)

spec/presenters/service_standard_presenter_spec.rb
thehenster: That one doesn't bother me personally but I've updated it to use 'returns'. :+1:

spec/forms/point_form_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+    edition = guide.editions.build
+    guide_form = described_class.new(guide: guide, edition: edition, user: User.new)
+    guide_form.save
+
+    expect(guide_form.errors.full_messages).to_not include("Topic section can't be blank")
+  end
+end
+
+RSpec.describe PointForm, "#slug_prefix" do
+  it "is /service-manual/service-standard" do
```
36degrees: And another.

PR #293, opened 2016-06-28
De-pin versions in Gemfile apart from Rails

thehenster:
To let bundler do it's job it's better to pin as few versions as possible.


36degrees:
Not worth locking even the major version?

thehenster:
Ollie has approved this verbally so merging..

PR #295, opened 2016-06-29
Save and publish the standard when publising a point

thehenster:
Because the draft of the standard can be overwritten (by saving a different point) before it is published we need to save and publish the standard when a point has just been published.

Don't merge yet.. I'm going to give this a decent poke on integration first.


36degrees:
👍

PR #298, opened 2016-06-29
Exclude the blank "" option from the page type filter drop down

36degrees:
compact only rejects nil values, but we get empty strings when doing `Guide.distinct(:type).pluck(:type)` so exclude them too.


thehenster:
:+1:

PR #299, opened 2016-06-29
Update Airbrake gem

thehenster:
The configuration api changed for version 5 of airbrake.

This works in integration.. I managed to get an error into the Errbit interface.


36degrees:
👍

PR #300, opened 2016-06-30
Remove the summary from editions

thehenster:
It turned out there was no need to maintain a separate field for the meta description and the short point text.

To avoid changing everything at once we are leaving the summary in the details hash in the schema. This can potentially be removed in favour of using the description field instead but there needs to be a decision on how we distinguish a guide from a point.

### DIFF
--- a/db/migrate/20160630082357_drop_editions_summary.rb
+++ b/db/migrate/20160630082357_drop_editions_summary.rb
@@ -1,0 +1,5 @@
+class DropEditionsSummary < ActiveRecord::Migration
+  def change
+    remove_column :editions, :summary
+  end
+end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -119,14 +119,14 @@ FactoryGirl.define do
     transient do
       title "An Unpublished Point"
       body "Some Body Text"
-      summary "A summary"
+      description "A summary"
     end
 
     editions {
       [
-        build(:edition, state: "draft", title: title, body: body, summary: summary),
-        build(:edition, state: "published", title: title, body: body, summary: summary),
-        build(:edition, state: "unpublished", title: title, body: body, summary: summary),
+        build(:edition, state: "draft", title: title, body: body, description: description),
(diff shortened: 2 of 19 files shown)


spec/factories.rb
```diff
@@ -119,14 +119,14 @@
     transient do
       title "An Unpublished Point"
       body "Some Body Text"
-      summary "A summary"
+      description "A summary"
```
36degrees: Worth making this transient in `unpublished_guide`, and then this can just inherit from `unpublished_guide` and just override the class? Guides have descriptions too right?

thehenster: Hmm. It looks like `:unpublished_point` isn't used.. I've removed it 🎉

thehenster: In a perfect world we'd have a reason to use this factory because we're not republishing the standard if we unpublish a point. But as there is no plan to change the number of points anytime soon and we'll get the unpublishing thing for free when we move to the links hash.. we get away with it.

db/migrate/20160630082357_drop_editions_summary.rb
```diff
@@ -0,0 +1,5 @@
+class DropEditionsSummary < ActiveRecord::Migration
+  def change
+    remove_column :editions, :summary
```
36degrees: When do migrations run? Before or after the code change is deployed?

thehenster: They run after.

thehenster: The code gets checked out, the migrations are run and then the app instances are restarted.

36degrees:
Yep, LGTM… 👍

thehenster:
I was going to break it into a separate PR.. but it's not worth it. I've added the `show_description` boolean to this PR.

This and https://github.com/alphagov/service-manual-frontend/pull/31 and https://github.com/alphagov/govuk-content-schemas/pull/336/files are now in sync.

PR #301, opened 2016-07-01
Remove the group contents from the topic payload

thehenster:
It seems as though it's not used in the frontend anymore.


36degrees:
:+1:

PR #302, opened 2016-07-01
Factories cleanup

36degrees:
Feedback please.

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -2,12 +2,52 @@ FactoryGirl.define do
   factory :guide do
     transient do
       title "Example Guide"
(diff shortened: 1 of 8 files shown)


thehenster:
👏  💯 

Could we add that example for how to create a bunch of guides with the same guide community / content owner?

And is the plan to remove the uses of the legacy factories and then the factories themselves after this is merged?

Merge when ready

36degrees:
Yes, a future PR will remove the calls to the 'legacy' factories and the legacy definitions. Will probably also try and nail the naming of the traits (`:with_review_requested_edition`, `:with_previously_published_edition`).

PR #303, opened 2016-07-01
Exclude unpublished guides

36degrees:


spec/models/guide_spec.rb
```diff
@@ -129,20 +129,29 @@
...
+        content_owner_id: guide_community.id
+      })
+      published_guide = create(:guide, :with_published_edition,
+        slug: "/service-manual/topic-name/2",
+        edition: {
+          content_owner_id: guide_community.id
+        }
+      )
+
+      expect(Guide.with_published_editions.to_a).to eq [guide_community, published_guide]
```
thehenster: Both of these are preexisting..

This is a class method so the describe block text should read `".with_published_editions"`, ie with a dot rather than a hash.

The `to_a` isn't necessary as rspec-rails is smart enough to load the records and compare the arrays rather than tell you that an Array isn't an ActiveRecord::Relation. Also, maybe use `#match_array` rather than `#eq` if we don't care about the order of the results.

spec/models/guide_spec.rb
```diff
@@ -129,20 +129,29 @@
...
+      guide_community = create(:guide_community, :with_published_edition)
       guide = create(:guide,
         slug: "/service-manual/topic-name/2",
-        editions: [
-          build(:published_edition),
-          build(:published_edition),
-        ],
-                    )
-      expect(Guide.with_published_editions.to_a).to eq [guide]
+        states: [:draft, :review_requested, :ready, :published] * 2,
```
thehenster: Neat :+1:

spec/models/guide_spec.rb
thehenster: (you probably know this but order isn't guaranteed in postgres without an `ORDER BY`)

spec/models/guide_spec.rb
```diff
@@ -129,20 +129,29 @@
...
+        edition: { content_owner_id: guide_community.id }
+      )
+
+      expect(Guide.with_published_editions.to_a).to eq [guide_community, guide]
```
thehenster: Same to_a and match_array suggestion here.

thehenster:
Minor comments only - LGTM :+1:

thehenster:
Nice to see the new factories coming in 👯

36degrees:
@thehenster re-review?

PR #305, opened 2016-07-04
Various validation error fixes

36degrees:
See individual commits for details.


spec/features/topic_spec.rb
```diff
@@ -114,6 +114,7 @@
 
     within '.full-error-list' do
       expect(page).to have_content "Path must be present and start with '/service-manual/'"
+      expect(page).to have_content "Title can't be blank"
```
thehenster: It'd be nice to test this in the model spec. There _should_ be similar assertions on the existing validations in there..

app/assets/stylesheets/topics.scss
```diff
@@ -1,4 +1,4 @@
-.topics {
+.topic-section-list {
```
thehenster: :+1:

How about renaming the file to topic-section-list.scss? I like it when the block name is reflected in the filename. We follow that pattern in service-manual-frontend. Not that important so do what feels right.

app/views/topics/_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="topics">
+<div class="row">
```
thehenster: Good removal of that topics class too :+1:

thehenster:
LGTM

app/assets/stylesheets/topics.scss
36degrees: This hasn't been marked as outdated, but has been done.

PR #307, opened 2016-07-05
Remember slug edits

36degrees:
I've also had a go at tidying up the slug generation feature spec, so it's worth reviewing this as two separate commits for clarity.

### DIFF
--- a/spec/features/slug_generation_spec.rb
+++ b/spec/features/slug_generation_spec.rb
@@ -24,29 +24,33 @@ RSpec.describe 'Generating slugs', type: :feature, js: true do
 
     context 'when the user has manually edited the slug' do
       it 'does not update the slug or final url when you change the title' do
-        guide = create(:guide, :with_draft_edition, :with_topic_section)
-        visit edit_guide_path(guide)
+        topic = create(:topic, path: '/service-manual/my-topic')
+        topic_section = create(:topic_section, topic: topic)
+        topic_section_label = "#{topic.title} -> #{topic_section.title}"
+
+        visit new_guide_path
 
         fill_in 'Slug', with: 'something'
         fill_in 'Title', with: 'My Guide Title'
-        select 'Agile Delivery -> Topic Section Title', from: 'Topic section', exact: true
+
+        select topic_section_label, from: 'Topic section', exact: true
 
         expect(find_field('Slug').value).to eq 'something'
-        expect(find_field('Final URL').value).to eq "#{guide.topic.path}/something"
+        expect(find_field('Final URL').value).to eq '/service-manual/my-topic/something'
       end
 
       it 'remembers that the slug was edited when coming back to edit it' do
-        topic_section = create(:topic_section)
+        topic = create(:topic, path: '/service-manual/my-topic')
+        topic_section = create(:topic_section, topic: topic)
 
-        slug = 'my-custom-slug'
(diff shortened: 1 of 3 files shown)


spec/features/slug_generation_spec.rb
```diff
@@ -1,63 +1,51 @@
 require 'rails_helper'
-require 'capybara/rails'
-require 'gds_api/publishing_api_v2'
```
thehenster: :+1:

spec/features/slug_generation_spec.rb
```diff
@@ -1,63 +1,51 @@
...
+  context 'when the guide has not been published' do
+    it 'updates the slug and final url when changing the title or section' do
+      examples = {
+        'two words': 'two-words',
+        'slug--with-----hyphens': 'slug-with-hyphens',
+        '       space    slugs  ': 'space-slugs',
+        'other things !@#$%^&*()_-+=/\\': 'other-things',
+      }
+      topic_section = create(:topic_section)
+      topic_section_label = "#{topic_section.topic.title} -> #{topic_section.title}"
```
thehenster: I like using literals in assertions where I can. I find it easier to read and more likely to catch unseen changes. For example..

```
topic_section = create(:topic_section, title: "Getting stuff done")
topic_section_label = "Getting stuff done -> Getting stuff done"
```

thehenster: I see it's not actually used in an assertion.. but you get what I mean.

spec/features/slug_generation_spec.rb
```diff
@@ -1,63 +1,51 @@
...
-      'other things !@#$%^&*()_-+=/\\': "other-things",
-    }
-  end
-
-
-  describe "slug generation" do
-    context "guide that has not been published" do
-      let :guide do
-        create(:guide, :with_draft_edition, :with_topic_section)
+RSpec.describe 'Generating slugs', type: :feature, js: true do
```
thehenster: It looks like all the examples had js enabled individually but safer and better to enable it for the block :+1:

spec/features/slug_generation_spec.rb
```diff
@@ -34,6 +34,20 @@
...
+        slug = 'my-custom-slug'
+        guide = create(:guide, :with_draft_edition, slug: "#{topic_section.topic.path}/#{slug}")
+        topic_section.guides << guide
+
+        visit edit_guide_path(guide)
+
+        fill_in 'Title', with: 'A New Title'
+        expect(find_field('Slug').value).to eq slug
+        expect(find_field('Final URL').value).to eq "#{guide.topic.path}/#{slug}"
+      end
```
thehenster: I wrote a comment and deleted it as my internal pettiness alarm kicked in but just in case you like it..

I like using real/literal values in tests where I can and only use variables/interpolation where I have to. I find it easier to read and more resilient to regressions.. For this test I'd be tempted to..

```
topic = create(:topic, path: '/service-manual/topic')
topic_section = create(:topic_section, topic: topic)
```  

which would let me write assertions like..

```
expect(find_field('Final URL').value).to eq "/service-manual/topic/my-custom-slug"
```

It saves the reader some thinking in my opinion. The downside is that there is duplication of values in the test.. which personally doesn't bother me at all. Not an important thing so up to you!

thehenster:
The cleanup of the slug generation spec is good :+1:

36degrees:
@thehenster take another look at the tests? I've made them more explicit as suggested.

PR #308, opened 2016-07-06
Use Publishing API's unpublish endpoint when unpublishing guides

36degrees:
- [x] Wait for [gds-api-adaptors change](https://github.com/alphagov/gds-api-adapters/pull/533) so that we don't have to change the vhost we use for rummager.
- [x] Handle publishing api being unavailable
- [x] Work out what's going on with Rummager throwing 404s and everything breaking.


app/helpers/redirect_destination_helper.rb
```diff
@@ -1,5 +1,5 @@
-module SlugMigrationsHelper
-  def select_options
+module RedirectDestinationHelper
+  def redirect_destination_select_options
```
thehenster: Also.. all helpers are included when rendering views from all controllers these days.. so using a more specific name than `#select_options` is good too :+1:

app/controllers/unpublishes_controller.rb
```diff
@@ -1,30 +0,0 @@
...
-        old_path:   @redirect.old_path,
-        new_path: @redirect.new_path,
-      )
-      GuideSearchIndexer.new(@guide).delete
-      redirect_to root_path
-    else
-      render :new
-    end
-  end
-end
```
thehenster: Moving this to the GuidesController is cool, but it doesn't ultra-offend me in the previous place. I think you could argue it's creating an unpublish resource just as easily as a non-RESTful action on a guide.

But... I probably would have done what you've done. Just the "Unpublishers" word is horrible to look at by itself!

app/models/guide_manager.rb
```diff
@@ -40,6 +40,27 @@ def publish
...
+        )
+
+        GuideSearchIndexer.new(@guide).delete
+
+        ManageResult.new(true, [])
+      else
+        ManageResult.new(false, 'Could not save redirect')
+      end
+    end
+  end
```
thehenster: It'd be good to test this in guide_manager_spec.rb

app/controllers/guides_controller.rb
```diff
@@ -61,9 +61,15 @@ def confirm_unpublish
...
-    guide_manager.unpublish(redirect)
+    result = guide_manager.unpublish(redirect)
 
-    redirect_to edit_guide_path(guide), notice: "Guide has been unpublished"
+    if result.success?
+      redirect_to edit_guide_path(guide), notice: "Guide has been unpublished"
+    else
+      redirect_to unpublish_guide_path(guide), flash: {
+        error: "Guide could not be unpublished"
+      }
```
thehenster: It's common to re-render the template for the "new" action when something fails. eg. https://github.com/alphagov/service-manual-publisher/blob/master/app/controllers/guides_controller.rb#L78

The benefit is that form fields get repopulated for free (most of the time). In this case we're probably not loosing a lot.. but ya know.. it's a pattern people expect I guess.

spec/features/unpublishing_guides_spec.rb
```diff
@@ -0,0 +1,210 @@
...
+
+    it 'unpublishes the content, creating a redirect' do
+      visit unpublish_guide_path(guide)
+      select topic.path, from: "Redirect to"
+      click_button "Unpublish"
+
+      assert_publishing_api_unpublish(guide.content_id,
+        type: 'redirect',
+        alternative_path: '/service-manual/suitable-redirect'
+      )
```
thehenster: Nice to use these helpers.

app/models/guide_manager.rb
```diff
@@ -40,6 +40,27 @@ def publish
...
+        PUBLISHING_API.unpublish(guide.content_id,
+          type: 'redirect',
+          alternative_path: redirect.new_path
+        )
+
+        GuideSearchIndexer.new(@guide).delete
+
+        ManageResult.new(true, [])
+      else
+        ManageResult.new(false, 'Could not save redirect')
```
thehenster: Could the redirect errors be put in here?

It's best if, like the success version, the errors thing talks like an array.

36degrees:
I think this is ready to be re-reviewed, please?

app/models/redirect.rb
```diff
@@ -3,4 +3,13 @@ class Redirect < ActiveRecord::Base
...
   validates :new_path, presence: true
+
+  HUMANIZED_ATTRIBUTES = {
+    old_path: 'Redirect source',
+    new_path: 'Redirect destination'
+  }.freeze
+
+  def self.human_attribute_name(attr, options = {})
+    HUMANIZED_ATTRIBUTES[attr.to_sym] || super
+  end
```
thehenster: Have a gander at http://guides.rubyonrails.org/i18n.html#translations-for-active-record-models for an alternate way to do this.

spec/models/guide_manager_spec.rb
```diff
@@ -203,24 +194,139 @@ def stub_publishing_api_to_fail
...
+
+    it 'sends a notification to Airbrake' do
+      user = create(:user)
+      guide = create(:guide, :with_published_edition)
+
+      expect(Airbrake).to receive(:notify)
+
+      manager = described_class.new(guide: guide, user: user)
+      manager.unpublish_with_redirect('/service-manual/suitable-redirect')
+    end
```
thehenster: Well done for testing this :+1:

thehenster:
Looks good :+1:

36degrees:
👀 on those last two commits would be good please @thehenster

thehenster:
LGTM

PR #310, opened 2016-07-13
Make the service standard searchable

36degrees:
By indexing it in Rummager 💃 🎉

### DIFF
--- a/app/models/service_standard_search_indexer.rb
+++ b/app/models/service_standard_search_indexer.rb
@@ -1,0 +1,25 @@
+class ServiceStandardSearchIndexer
+  def initialize
+    @service_standard = ServiceStandardPresenter.new(Point.all).content_payload
+  end
+
+  def index
+    RUMMAGER_API.add_document(
+      # Pretend the service standard is a guide, as far as Rummager's
+      # concerned it might as well be.
+      'service_manual_guide',
+      @service_standard[:base_path],
+      'format':            'service_manual_guide',
+      'description':       @service_standard[:details][:introduction],
+      'indexable_content': @service_standard[:details][:body],
+      'title':             @service_standard[:title],
+      'link':              @service_standard[:base_path],
+      'manual':            '/service-manual',
+      'organisations':     ['government-digital-service'],
+    )
+  end
+
+  def delete
+    RUMMAGER_API.delete_content!(@service_standard[:base_path])
+  end
+end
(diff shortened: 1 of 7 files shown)


spec/models/service_standard_search_indexer_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+end
+
+RSpec.describe ServiceStandardSearchIndexer, '#delete' do
+  it 'deletes documents from rummager' do
+    expect(RUMMAGER_API).to receive(:delete_content!)
+      .with('/service-manual/service-standard')
+
+    described_class.new.delete
+  end
```
thehenster: Could we use https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/rummager.rb#L43-L52 ?

thehenster: And the add_document version for the test above?

app/models/service_standard_search_indexer.rb
```diff
@@ -0,0 +1,25 @@
...
+      'link':              @service_standard[:base_path],
+      'manual':            '/service-manual',
+      'organisations':     ['government-digital-service'],
+    )
+  end
+
+  def delete
+    RUMMAGER_API.delete_content!(@service_standard[:base_path])
+  end
+end
```
thehenster: Yeah, makes sense to use the presenter.

thehenster:
Just commented on using the gds-api-adapter test helpers. Looks good though.

It might be nice to separate the guide indexing task into it's own task so guides can be indexed independently of the service standard? Then the `rake rummager:index` task could have each of the content type indexing tasks as dependencies.

36degrees:
Yeah, I started going down the route of using the rummager helpers but at the minute I'd have to duplicate some of the changes made in #308, including waiting for a new release of the `gds-api-adaptors` change. So I held off on that. I think I'll try and get that merged today, rebase on top and fix the tests to use the helper.

36degrees:
:tada: Re-review please @thehenster

spec/models/guide_search_indexer_spec.rb
```diff
@@ -1,47 +1,52 @@
...
+  it 'does not attempt to index a guide if it has no live editions' do
     guide = create(:guide)
 
-    expect(rummager_api).to_not receive(:add_document)
+    described_class.new(guide).index
 
-    described_class.new(guide, rummager_api: rummager_api).index
+    assert_not_requested(
+      :post,
+      %r{#{GdsApi::TestHelpers::Rummager::RUMMAGER_ENDPOINT}/documents}
```
thehenster: It would be minorly more resilient to assert it doesn't call any Rummager URL. Not a biggie though. Whichever you prefer.

thehenster:
Looks good

PR #311, opened 2016-07-15
Do not send public updated at

thehenster:
https://trello.com/c/QzWdrdgO

This includes a bit of clean up for the topic presenter test.

The complication was that we were using `updated_at` on the frontend which is fixed by this: https://github.com/alphagov/service-manual-frontend/pull/42

### DIFF
--- a/spec/factories/topic_section.rb
+++ b/spec/factories/topic_section.rb
@@ -1,7 +1,26 @@
 FactoryGirl.define do
+  # Example Usage
+  #
+  # Create a topic section with the title Hello World
+  # -> create(:topic_section, title: "Hello World")
+  #
+  # Create a topic section with associated guides
+  # -> create(:topic_section, guides: [create(:guide), create(:guide)])
+
   factory :topic_section do
+    transient do
+      guides []
+    end
+
     topic
     title "Topic Section Title"
     description "Topic Section Description"
+
+    after(:build) do |topic_section, evaluator|
+      evaluator.guides.each do |guide|
+        topic_section_guide = TopicSectionGuide.new(guide: guide, topic_section: topic_section)
+        topic_section.topic_section_guides << topic_section_guide
+      end
+    end
   end
 end
--- a/spec/presenters/guide_presenter_spec.rb
+++ b/spec/presenters/guide_presenter_spec.rb
@@ -41,8 +41,13 @@ RSpec.describe GuidePresenter do
       )
     end
 
-    it "doesn't contain public_updated_at so that it represents the published at time" \
-     " on the frontend" do
+    it "omits public_updated_at" do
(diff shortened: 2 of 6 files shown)


spec/factories/topic_section.rb
```diff
@@ -1,7 +1,26 @@
 FactoryGirl.define do
+  # Example Usage
```
36degrees: 👍

spec/presenters/topic_presenter_spec.rb
```diff
@@ -1,67 +1,77 @@
...
-        update_type: "minor",
-        phase: "beta",
-        publishing_app: "service-manual-publisher",
-        rendering_app: "service-manual-frontend",
-        format: "service_manual_topic",
-        locale: "en",
-        base_path: "/service-manual/test-topic"
-      )
-    end
+  it "sets visually_expanded" do
```
36degrees: visually_collapsed

spec/presenters/guide_presenter_spec.rb
```diff
@@ -41,6 +41,11 @@
       )
     end
 
+    it "doesn't contain public_updated_at so that it represents the published at time" \
```
36degrees: My brain struggles with this sentence. Might be the negation combined with the 'so that' clause? Not sure what would be better, but maybe changing 'doesn't contain' to e.g. 'omits' might start to clear things up?

thehenster:
@36degrees I've reworded the test and moved it to a shared place. Is it easier to read?

spec/presenters/guide_presenter_spec.rb
```diff
@@ -28,15 +28,18 @@
       expect(presenter.content_payload).to be_valid_against_schema('service_manual_guide')
     end
 
+    describe "common service manual draft payload" do
+      let(:payload) { presenter.content_payload }
+
+      include_examples "common service manual draft payload"
```
36degrees: Unless I'm missing a subtle difference between `include_examples` and `it_behaves_like "[a…]"` I find the latter works a lot better within the 'english as code' context.

But that's probably also just because I'm more used to it, happy to be persuaded otherwise.

36degrees:
Relatively minor personal preference thing, otherwise looks good.

spec/presenters/guide_presenter_spec.rb
thehenster: I went off `it_behaves_like` because it doesn't always make sense depending on the text. eg. `it_behaves_like "catch http errors"`. Whereas `include_examples` always works because you don't even try to read it like a sentence..

it_behaves_like would read sensibly if it was "a common service manual draft payload" so we could change it.

Generally I find it a losing battle..

PR #312, opened 2016-07-18
Remove some requires

thehenster:
Maybe some of these made more sense if we were requiring spec_helper in the past.


36degrees:
Yep.

PR #314, opened 2016-07-19
Remove unpublished guides from dropdowns

thehenster:
https://trello.com/c/jKTudCQA/364-remove-unpublished-guides-from-relevant-dropdowns

### DIFF
--- a/app/helpers/redirect_destination_helper.rb
+++ b/app/helpers/redirect_destination_helper.rb
@@ -1,7 +1,7 @@
 module RedirectDestinationHelper
   def redirect_destination_select_options
     guide_select_options = Guide
-      .with_published_editions
+      .live
       .order(:slug).pluck(:slug)
       .map { |g| [g, g] }
 
--- a/spec/helpers/guide_helper_spec.rb
+++ b/spec/helpers/guide_helper_spec.rb
@@ -2,15 +2,26 @@ require 'rails_helper'
 
 RSpec.describe GuideHelper, '#guide_community_options_for_select', type: :helper do
   it 'returns an array of options for a select tag in alphabetical order' do
-    second = create(:guide_community, editions: [build(:edition, title: 'Banana', content_owner: nil)])
-    first = create(:guide_community, editions: [build(:edition, title: 'Apple', content_owner: nil)])
-    third = create(:guide_community, editions: [build(:edition, title: 'Cucumber', content_owner: nil)])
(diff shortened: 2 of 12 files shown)


spec/helpers/guide_helper_spec.rb
```diff
@@ -2,9 +2,9 @@
 
 RSpec.describe GuideHelper, '#guide_community_options_for_select', type: :helper do
   it 'returns an array of options for a select tag in alphabetical order' do
-    second = create(:guide_community, editions: [build(:edition, title: 'Banana', content_owner: nil)])
-    first = create(:guide_community, editions: [build(:edition, title: 'Apple', content_owner: nil)])
-    third = create(:guide_community, editions: [build(:edition, title: 'Cucumber', content_owner: nil)])
+    second = create(:guide_community, :with_published_edition, title: 'Banana')
+    first = create(:guide_community, :with_published_edition, title: 'Apple')
+    third = create(:guide_community, :with_published_edition, title: 'Cucumber')
```
36degrees: Draft should be included too right? Worth having a draft in the test fixtures?

thehenster: It's tested in https://github.com/alphagov/service-manual-publisher/pull/314/files#diff-1e73b14f1ae43ea29a378b4874d36140R344 but it doesn't hurt to replace one of these with a draft.. I'll update it.

app/helpers/redirect_destination_helper.rb
```diff
@@ -1,7 +1,7 @@
 module RedirectDestinationHelper
   def redirect_destination_select_options
     guide_select_options = Guide
-      .with_published_editions
+      .live
```
36degrees: Mainly out of curiosity… we have leading periods in this method and trailing in the guide_helper method. Do we have a standard way of doing this, and should we be following it?

36degrees:
LGTM 👍

app/helpers/redirect_destination_helper.rb
thehenster: Sounds like an accident. Probably mine. We agreed on this project to use leading dots so that's what they should be.

Neither the alphagov or the bbatsov style guide have a preference although the latter talks about it: https://github.com/bbatsov/ruby-style-guide#consistent-multi-line-chains

My addition to the argument is that trailing dots let you copy and paste stuff into irb/console. With a leading dot it doesn't know if there is another chained method coming and so usually errors.

PR #317, opened 2016-07-20
Fix deleting topic sections

36degrees:
When I removed the topics class from the body of the topic page section deletion stopped working. We didn't notice because it wasn't tested.

So add a test, and then fix it.


thehenster:
:+1:

PR #319, opened 2016-07-21
Allow users to reorder topic sections and guides

36degrees:
The UI for this was previously built but apparently never actually worked (or at least, not since it was refactored).

This adds `position` hidden fields for each topic section and guide. For the overall topic section list, and each list of guides within each section we create an instance of [Dragula](https://github.com/bevacqua/dragula) and hook it up such that whenever a drag event occurs we iterate over the `position` fields in the respective list and re-number them in the order they appear.


spec/factories/topic_section.rb
```diff
@@ -15,6 +15,7 @@
     topic
     title "Topic Section Title"
     description "Topic Section Description"
+    position 0
```
thehenster: Is this intentionally different to the acts_as_list default of 1? Maybe it doesn't matter.. A new section will have a position of `max + 1` or `0`.. even if the max is 0 it will still order correctly..

spec/features/reordering_topic_sections.rb
```diff
@@ -0,0 +1,69 @@
...
+  def drag_topic_section_above(dragged_section_title, destination_section_title)
+    handle_for_topic_section(dragged_section_title).drag_to(
+      handle_for_topic_section(destination_section_title)
+    )
+  end
+
+  def sections_in_order
+    all('.list-group-item input[placeholder="Heading Title"]').map(&:value)
+  end
+end
```
thehenster: Good tests :+1:

app/models/topic_section_guide.rb
```diff
@@ -1,10 +1,21 @@
...
+    where(topic_section_id: topic_section_id)
+  }
+
+  def default_position_to_next_in_list
+    self.position ||= next_position_in_list
+  end
+
+  def next_position_in_list
+    (TopicSectionGuide.within_topic_section(self.topic_section_id).maximum(:position) || 0) + 1
+  end
```
thehenster: Could this be a private method?

thehenster: Maybe the same goes for `#default_position_to_next_in_list`

app/models/topic.rb
```diff
@@ -32,8 +32,16 @@ def guide_content_ids
...
 
+  def add_section
+    self.topic_sections.build(position: next_position_in_list)
+  end
+
 private
 
+  def next_position_in_list
+    (TopicSection.belonging_to_topic(self.id).maximum(:position) || 0) + 1
+  end
```
thehenster: The self in `self.id` isn't necessary here. As it won't be able to find a local variable called `id` it will look for the method `#id` anyway.

thehenster: Very optional but if this was broken into two methods `#highest_position` and `#next_position` (or something like that) it would avoid the `|| 0) + 1` sequence which is a little COMPUTER.

app/views/topics/_form.html.erb
```diff
@@ -37,9 +37,10 @@
       <div class="well">
         <h3>Grouping</h3>
         <ul class="list-group topic-section-list js-topic-section-list">
-          <%= f.fields_for :topic_sections do |topic_section_form| %>
+          <%= f.fields_for :topic_sections, f.object.topic_sections.sort_by(&:position) do |topic_section_form| %>
```
thehenster: There is probably a N+1 to be squashed here with an `.includes(:guides)`. It may well be needed in the controller too where we build new topic sections.

The world will keep turning without fixing it off course.. Performance probably isn't a big issue anywhere in this app :)

thehenster: Imagine this comment is just below.. where we're iterating over the guides.

app/controllers/topics_controller.rb
```diff
@@ -76,6 +76,14 @@ def respond_for_topic_publication(publication, opts = {})
     end
   end
 
+  def next_position_in_list(topic)
+    highest_position_in_list(topic) + 1
+  end
+
+  def highest_position_in_list(topic)
+    topic.topic_sections.map(&:position).max || 0
```
thehenster: Didyaknow `topic.topic_sections.max_by(&:position)` would do the same? No need to change of course. They're virtually identical.

spec/features/reordering_guides.rb
```diff
@@ -0,0 +1,93 @@
...
+    section2.guides << create(:guide, title: "Guide A")
+
+    visit edit_topic_path(topic)
+    drag_guide_above("Guide A", "Guide B")
+
+    expect(guides_within_section("Section 1")).to eq ["Guide B"]
+    expect(guides_within_section("Section 2")).to eq ["Guide A"]
+  end
+
+  it 'remembers order changes when you add a heading' do
```
thehenster: I'm sure you have your reasons - why is this worth testing?

thehenster: Ah, I think I've got it. It's because we're not persisting at the add heading stage. Makes sense.

spec/features/reordering_topic_sections.rb
```diff
@@ -0,0 +1,102 @@
...
+  end
+
+  context 'when editing an existing topic' do
+    let(:topic) { create(:topic) }
+
+    before do
+      topic.topic_sections << create(:topic_section, title: "Section B", topic: topic, position: 1)
+      topic.topic_sections << create(:topic_section, title: "Section A", topic: topic, position: 1)
+      topic.topic_sections << create(:topic_section, title: "Section C", topic: topic, position: 1)
+    end
```
thehenster: Why do all these have a position of 1?

app/models/topic_section.rb
```diff
@@ -1,8 +1,11 @@
...
   has_many :guides, through: :topic_section_guides
-  accepts_nested_attributes_for :guides
 
   has_many :topic_section_guides, -> { order(position: :asc) }, dependent: :destroy
-  acts_as_list scope: :topic
+  accepts_nested_attributes_for :topic_section_guides
+
+  scope :belonging_to_topic, ->(topic_id) {
+    where(topic_id: topic_id)
+  }
```
thehenster: I can't see where this is getting used. A leftover from a refactor?

thehenster:
Good approach :+1: It's a bit of a tricky one with needing to keep persisted and unpersisted topic sections along with their related guides all in order.

I think I would personally prefer the topic sections to be assigned to an instance variable in the controller and used in the `fields_for` call in the view. It's not immediately obvious when looking at the view that the collection is already loaded and manipulated. Also, if that N+1 exists then you'll only need to do an includes in the controller.

I gave it a poke on my local. Seems to do all the right stuff.

36degrees:
> I think I would personally prefer the topic sections to be assigned to an instance variable in the controller and used in the fields_for call in the view. It's not immediately obvious when looking at the view that the collection is already loaded and manipulated.

I think I had that originally, but disliked the fact that it had to be done in four places in the controller (new, create, edit, update). But reaching out from the view isn't great either.

PR #320, opened 2016-07-25
Remove duplicate elements in redirect select options

deborahchua:
In the RedirectDestinationHelper, it was not necessary to map the array when .order.pluck already does this. This has now been removed.


36degrees:
🎉

PR #321, opened 2016-07-25
Omit 'editions is invalid' error message

36degrees:
It doesn't mean anything to the end user, so let's not include it when promoting errors from the guide model.

### DIFF
--- a/app/forms/base_guide_form.rb
+++ b/app/forms/base_guide_form.rb
@@ -128,8 +128,9 @@ private
   end
 
   def promote_errors_for(model)
+    ignored_attributes = [:editions]
     model.errors.each do |attrib, message|
-      errors.add(attrib, message)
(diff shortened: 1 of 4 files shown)


app/forms/base_guide_form.rb
```diff
@@ -128,8 +128,9 @@ def next_author_id
   end
 
   def promote_errors_for(model)
+    ignored_attributes = [:editions]
     model.errors.each do |attrib, message|
-      errors.add(attrib, message)
+      errors.add(attrib, message) unless ignored_attributes.include? attrib
```
thehenster: How about passing in which attribute(s) to ignore as a method argument? Something like..

```
def promote_errors_for(model, opts = {})
  ignored_attributes = opts.fetch(:ignored_attributes, [])
```

thehenster:
:+1:

PR #322, opened 2016-07-25
Include change history in content payload for guides

36degrees:


36degrees:
For context, this was [previously attempted](https://github.com/alphagov/service-manual-publisher/pull/209) and then [reverted](https://github.com/alphagov/service-manual-publisher/pull/222), so this is a second attempt…

app/models/edition.rb
```diff
@@ -18,10 +18,13 @@ class Edition < ActiveRecord::Base
   scope :most_recent_first, -> { order('created_at DESC, id DESC') }
   scope :which_update_the_frontend, -> { where(state: STATES_THAT_UPDATE_THE_FRONTEND) }
 
+  scope :major, -> { where(update_type: 'major') }
+  scope :minor, -> { where(update_type: 'minor') }
```
thehenster: Did we need this in the end?

app/presenters/guide_presenter.rb
```diff
@@ -50,6 +50,7 @@ def details
     details_hash = {
       body: govspeak_body.to_html,
       header_links: level_two_headers,
+      change_history: GuidePresenter::ChangeHistoryPresenter.new(guide, edition).change_history
```
thehenster: There shouldn't be a need for `GuidePresenter::` in the above because `ChangeHistoryPresenter` is in the namespace.

spec/models/edition_spec.rb
```diff
@@ -86,16 +86,16 @@
...
+      it "does not allow empty reason_for_change when the update_type is 'major'" do
+        edition = build(:edition, update_type: "major", reason_for_change: "")
         edition.valid?
-        expect(edition.errors.full_messages_for(:change_note)).to eq ["Change note can't be blank"]
+        expect(edition.errors.full_messages_for(:reason_for_change)).to eq ["Reason for change can't be blank"]
       end
 
-      it "allows empty change_note when the update_type is 'minor'" do
-        edition = build(:edition, update_type: "minor", change_note: "")
+      it "allows empty reason_for_change when the update_type is 'minor'" do
```
thehenster: This is just part of the rename but it might be good to test that change_note can be empty on minor too? And probably can't be empty on major too.

db/migrate/20160729100003_rename_editions_change_note_and_change_summary.rb
```diff
@@ -0,0 +1,6 @@
+class RenameEditionsChangeNoteAndChangeSummary < ActiveRecord::Migration
+  def change
+    rename_column :editions, :change_note, :reason_for_change
+    rename_column :editions, :change_summary, :change_note
+  end
+end
```
thehenster: Very sensible rename if I haven't said that already.

spec/models/edition_spec.rb
```diff
@@ -234,15 +234,15 @@
     end
   end
 
-  describe "#change_note_html" do
+  describe "#reason_for_change_html" do
```
thehenster: I can't see where this is getting used..

It only occured to me to search for it as we're applying simple_format to the reason for change in the frontend and it didn't make sense to have html tags in there.

PR #325, opened 2016-08-04
[DO NOT MERGE] Topic controller juggle

thehenster:


app/controllers/topics_controller.rb
```diff
@@ -9,60 +12,52 @@ def new
...
 
-    if params[:publish]
-      publish
+    if params[:add_heading]
+      add_heading(@topic)
+    elsif params[:publish]
+      respond_for_topic_publication publish(@topic), notice: "Topic has been published"
     else
-      save_draft
+      respond_for_topic_publication save_draft(@topic), notice: "Topic has been updated"
```
36degrees: I find it weird that we're using `respond_for_topic_publication` to handle the output of a draft being saved, maybe it needs to be renamed?

thehenster: Yeah, I agree. I'll add a commit.

PR #327, opened 2016-08-09
First edition is major

thehenster:
https://trello.com/c/fq2gTSIH

The publishing-api requires that the first edition has an update_type of 'major'.

Our content editors had the requirement that they did not want to give a change_note or reason_for_change for the first edition.

We automatically set the first version's change_note to "Guidance first published" and reason_for_change as an empty string.

The form fields have been removed for the first version so the user can't edit them anymore. The existing validations on change_note and reason_for_change have been changed and we've added a validation to ensure the first edition is minor too, partially for to protect against something changing later, and partially to ensure more real-life data in tests.


36degrees:
All looks sensible to me, yep.

PR #328, opened 2016-08-11
Scope cleanup

thehenster:
Just a little cleanup. The edition scopes should live in the Edition model and renamed the methods to be a little clearer.


36degrees:
Makes sense, yep.

PR #330, opened 2016-08-11
Edition policy methods

thehenster:
I noticed the guard clauses on the policy-like methods on the `Edition` class, as their tests, were weird. I've rewritten the tests, fixed them and then ended up moving them to a `EditionPolicy` object.

`app/views/editions/_actions.html.erb` is weird too. The logic regarding which buttons to display is mixed between the model and the view. I'd like to tackle it later but stopping here to avoid doing too much at once.


app/views/editions/_actions.html.erb
```diff
@@ -2,41 +2,40 @@
   <div class="panel-body">
     <div class="btn-toolbar" role="toolbar">
 
-      <% if edition.persisted? %>
+      <% edition_policy = EditionPolicy.new(current_user, edition) %>
```
36degrees: Would this make more sense being created in the controller?

thehenster: At the moment it feels like a view layer concern to me. It's just involved with displaying buttons or not. This kind of thing can go into helpers but I don't think it would be a big win at the mo.

The controller would definitely be the right place for this thing if we were implementing these conditions server side.. Which we should be doing! :)

36degrees:
👍

PR #333, opened 2016-08-16
Use heredocs with strip_heredoc to avoid outdenting multiline strings in tests

36degrees:
💃


thehenster:
Cool - I didn't know that method existed. I was waiting for `<<~`.

PR #334, opened 2016-08-17
Move topic section validation

thehenster:
The goals of this were to:

a) Allow all validations to be seen at once when saving an invalid guide
b) Put all Guide validations in one place

The tests now require more topics and topic sections to be created via factories but hopefully the benefit is that the state of the data in any one test is closer to real life.

I think now that the GuideForm has been reduced it might make sense to made `BaseGuideForm`, `GuideForm` and `PointForm` just `GuideForm` again (with a few conditionals inside it). The conditionals might be less of a brain drain than the class hierarchy.

### DIFF
--- a/app/forms/guide_form.rb
+++ b/app/forms/guide_form.rb
@@ -1,12 +1,7 @@
 class GuideForm < BaseGuideForm
   attr_accessor :topic_section_id
 
-  validates_presence_of :topic_section_id, if: :requires_topic?
   validate :topic_cannot_change
-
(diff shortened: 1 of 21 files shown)


36degrees:
Tests are failing because of [changes in govuk-content-schema](https://github.com/alphagov/govuk-content-schemas/pull/374) – fix in #335

36degrees:
Rebased now that #335 has been merged into master.

thehenster:
:+1:

PR #335, opened 2016-08-18
Replace deprecated format field with schema_name, document_type

36degrees:
The `format` field has been removed from the list of allowed attributes in the schema definition. We now need to send both `schema_name` and `document_type` instead.

Related to https://github.com/alphagov/govuk-content-schemas/pull/374


mobaig:
👍

36degrees:
Thanks, @mobaig!

PR #337, opened 2016-08-22
Publish email alert signups for topics

36degrees:
- When saving the draft for a topic, create a draft for an `email_alert_signup` for that topic.
- The `email_alert_signup` defines a confusingly named `subscriber_list` which is really a 'matcher' against the `service_manual_topics` link – when new major editions are created that match, an email will be sent to everyone subscribed.
- When publishing the topic, publish the corresponding `email_alert_signup` as well.
- Link the topic to the signup
- Include the topic when patching links for guides, so that the link exists when the guide is published.


app/models/guide.rb
```diff
@@ -9,6 +9,8 @@ class Guide < ActiveRecord::Base
   has_many :editions, dependent: :destroy
   has_many :topic_section_guides, autosave: true
 
+  has_one :topic, through: :topic_section_guides
```
thehenster: There is an instance method, `#topic`, which will clash with this. Did we need some of the association-y properties?

app/models/guide_tagger_job.rb
```diff
@@ -1,16 +0,0 @@
...
-    end
-  end
-
-  def perform(guide_content_id:, topic_content_id:)
-    PUBLISHING_API.patch_links(
-      guide_content_id,
-      links: { service_manual_topics: [topic_content_id] }
-    )
-  end
-end
```
thehenster: 🎂

spec/presenters/guide_presenter_spec.rb
```diff
@@ -102,27 +119,31 @@
...
     end
-  end
-end
 
-RSpec.describe GuidePresenter, "for a Point" do
-  it "includes the service standard as a parent in the links" do
-    edition = create(:edition)
-    point = create(:point, editions: [edition])
+    context 'for a guide community' do
+      let(:guide) { create(:guide_community) }
```
thehenster: Overriding `let`s can be confusing. Starting a new `describe` block is nice way of starting fresh. It's another trick for making the reader jump around less.. If you're still keen to use nesting then it's best to move the initial `let` into it's own context.

db/migrate/20160816150906_add_email_alert_signup_content_ids_to_topics.rb
```diff
@@ -0,0 +1,9 @@
+class AddEmailAlertSignupContentIdsToTopics < ActiveRecord::Migration
+  def change
+    add_column :topics, :email_alert_signup_content_id, :string
+
+    Topic.all.each do |topic|
```
thehenster: Using real models in migrations can be dangerous. `Topic` can change over time and callbacks and expected behaviour can change. This isn't a very risky one but it's generally good practise to either a) use raw SQL b) make a fake model. The fake model would look like..

```
class FakeTopic < ActiveRecord::Base
  self.table_name = "topics"
end

FakeTopic.all do |topic|
# etc
```

Here is another example from this app where we needed the associations too: https://github.com/alphagov/service-manual-publisher/blob/master/db/migrate/20160225130400_drop_content_owners.rb

thehenster:
Generally looks good :+1:

I think leaving the refactor of  `TopicPublisher` until later was the right thing to do. When we add drafts/editions to the topics is the right time for that.

app/models/guide.rb
36degrees: I failed to notice the existing method, I just knew I wanted a nice way of getting from Guide to its topic without having to traverse topic_section_guides. As it happens, the existing topic method does what I want so… that's all good then. Removed this and fixup-ed.

db/migrate/20160816150906_add_email_alert_signup_content_ids_to_topics.rb
36degrees: Updated and fixup-ed.

36degrees: Also added an index with a unique constraint for that column. We don't appear to have one for content_id but it felt like it should have one…

36degrees:
@thehenster Could you take a look at the last commit? I've added a task to republish all the topics so that we can easily add links to the email alert signups.

36degrees:
Discussed in real life – the topic republishing task needs a warning message that it will publish things that might not have been previously published, yada yada yada etc etc.

spec/presenters/guide_presenter_spec.rb
```diff
@@ -1,49 +1,35 @@
...
-        reason_for_change: "We added this guide so we can test the presenter"
-      },
-      slug: '/service-manual/test-topic/the-title',
-      topic: topic
-    )
-  end
-
-  let(:edition) { guide.latest_edition }
-
-  let(:presenter) { described_class.new(guide, edition) }
```
thehenster: :+1: for replacing this with duplication in the examples.

thehenster:
:+1:

(just the warning tape around the rake task we chatted about)

36degrees:
@thehenster Please check the latest commit.

PR #338, opened 2016-08-22
Redirect tech code of practise

thehenster:
This is a one off task. At the moment we don't plan to redirect to pages outside the service manual on a regular basis.


36degrees:
Should you update the respective slug migrations so that we know they're 'done'?

thehenster:
Yes - thanks.

36degrees:
Yarp.

PR #339, opened 2016-08-22
Revert topic section validation

thehenster:
Revert "Merge pull request #334 from alphagov/move-topic-section-validation"

While the validation has moved it isn't displaying the validation error. It's going to be an easy fix but we want to deploy and don't this holding it up.


36degrees:
Yep.

PR #340, opened 2016-08-23
Move the topic section validation to Guide

thehenster:
This is the rebirth of https://github.com/alphagov/service-manual-publisher/pull/334


36degrees:
Makes sense to me.

PR #341, opened 2016-08-23
Build the service standard points from the expanded links

thehenster:
Use link expansion to avoid saving and publishing the service standard every time a point changes.

Frontend PR: https://github.com/alphagov/service-manual-frontend/pull/50


36degrees:
👍

PR #343, opened 2016-09-14
Allow the Homepage to be published

36degrees:
This updates the topic presenter to include a link to the homepage as a `parent` if the topic is set to be included on the homepage (the communities topic won't be, and we need a way to filter that out)

We don't currently have a good way to publish the homepage, so the HomepagePresenter only exists to define a content id for the homepage. I'm not sure of the implications of creating links to things that don't exist yet, so that's why this PR is marked as do not merge.

### DIFF
--- a/app/presenters/topic_presenter.rb
+++ b/app/presenters/topic_presenter.rb
@@ -36,6 +36,7 @@ class TopicPresenter
         linked_items: topic.guides.map(&:content_id),
         content_owners: content_owner_content_ids,
         organisations: [ServiceManualPublisher::GDS_ORGANISATION_CONTENT_ID],
+        parent: parents
       }
     }
   end
@@ -43,6 +44,15 @@ class TopicPresenter
 private
 
   attr_reader :topic
+
+  def parents
+    if topic.include_on_homepage?
+      [HomepagePresenter::HOMEPAGE_CONTENT_ID]
+    else
+      []
+    end
+  end
+
 
   def groups
     topic.topic_sections.map do |topic_section|
--- a/lib/tasks/migrate_homepage.rake
+++ b/lib/tasks/migrate_homepage.rake
@@ -1,0 +1,38 @@
+require 'gds_api/publishing_api'
+require 'highline'
+
+desc "Migrate the homepage"
+task migrate_homepage: :environment do
+  cli = HighLine.new
+  cli.say cli.color(
+    "Publishing the homepage will make the old service manual inaccessible. Continue?",
+    :red
+  )
+  exit unless cli.agree "Are you sure you wish to continue?"
+
+  publishing_api_v1 = GdsApi::PublishingApi.new(
+    Plek.new.find('publishing-api'),
+    bearer_token: ENV['PUBLISHING_API_BEARER_TOKEN'] || 'example'
+  )
+
+  # Take ownership of the path reservation
+  publishing_api_v1.put_path('/service-manual',
+    publishing_app: 'service-manual-publisher',
+    override_existing: true
+  )
+
+  homepage = HomepagePresenter.new
+  
+  # Save and publish the homepage
+  PUBLISHING_API.put_content(
+    homepage.content_id,
+    homepage.content_payload
+  )
+
+  PUBLISHING_API.publish(
+    homepage.content_id,
+    "major"
+  )
+
+  puts "Done."
+end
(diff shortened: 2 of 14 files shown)


spec/controllers/topics_controller_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+        stub_any_rummager_post
+        topic = create(:topic, path: '/service-manual/a-topic')
+
+        put :update, id: topic.id, publish: true
+
+        assert_rummager_posted_item(_id: '/service-manual/a-topic')
+      end
+    end
+  end
+end
```
thehenster: Why did you opt for writing controller tests rather than add to the feature spec?

spec/features/topic_spec.rb
```diff
@@ -11,6 +11,12 @@
     expect(page).to_not have_button('Publish')
   end
 
+  it 'allows you to choose whether the topic should appear on the homepage' do
+    visit new_topic_path
+
+    expect(page).to have_checked_field 'Include on homepage?'
+  end
```
thehenster: :+1:

spec/presenters/homepage_presenter_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      base_path: '/service-manual',
+      routes: [
+        { type: 'exact', path: '/service-manual' }
+      ]
+  end
+
+  it 'includes the rendering and publishing apps' do
+    payload = described_class.new.content_payload
+
+    expect(payload).to include \
```
thehenster: The `\` here is kinda neat. Let's give it a whirl for a bit. Brackets are the more normal option.

app/presenters/topic_presenter.rb
```diff
@@ -36,6 +36,7 @@ def links_payload
         linked_items: topic.guides.map(&:content_id),
         content_owners: content_owner_content_ids,
         organisations: [ServiceManualPublisher::GDS_ORGANISATION_CONTENT_ID],
+        parent: parents
```
thehenster: I still think doing this is our best option rather than tackling a custom link expansion.

We should probably add to the schema example: https://github.com/alphagov/govuk-content-schemas/blob/master/formats/service_manual_topic/frontend/examples/service_manual_topic.json

lib/tasks/migrate_homepage.rake
```diff
@@ -0,0 +1,38 @@
...
+require 'highline'
+
+desc "Migrate the homepage"
+task migrate_homepage: :environment do
+  cli = HighLine.new
+  cli.say cli.color(
+    "Publishing the homepage will make the old service manual inaccessible. Continue?",
+    :red
+  )
+  exit unless cli.agree "Are you sure you wish to continue?"
```
thehenster: Faaancy :)

lib/tasks/migrate_homepage.rake
```diff
@@ -0,0 +1,38 @@
...
+
+  publishing_api_v1 = GdsApi::PublishingApi.new(
```
thehenster: We're using v1 because of something to do with the path reservation yes? Could you remind me quickly.. I has forgot.

spec/controllers/topics_controller_spec.rb
36degrees: Erm.

Honestly, I have no idea. It probably seemed like a good idea at the time, but I wrote this a month ago. These can probably be removed, if they're not adding anything.

spec/presenters/homepage_presenter_spec.rb
36degrees: I like normal. Not sure why I decided to do this instead. Happy to bracket it up…

lib/tasks/migrate_homepage.rake
36degrees: [Yep](https://github.com/alphagov/publishing-api/blob/master/doc/api.md#put-pathsbase_path).

app/presenters/topic_presenter.rb
36degrees: [This is done](https://github.com/alphagov/govuk-content-schemas/pull/420).

36degrees:
@mcmillan would you mind taking another look at this? I've rebased it, but the last two commits are entirely new:

- Changed the route type for the homepage from exact to prefix, so any requests for anything starting with '/service-manual' come to our app. We'll therefore be able to e.g. see 404s in our logs.
- Realised we need to republish topics as part of migration so that the relationship with the homepage exists.

PR #350, opened 2016-10-17
Guide has button to view live version when published

deborahchua:
A 'View on website' button has been added which allows users to view the live version of the guide. It is always shown unless it has never been published.:w
The 'preview' button is only shown when a guide has been saved in a draft state.

### DIFF
--- a/app/models/guide.rb
+++ b/app/models/guide.rb
@@ -89,7 +89,7 @@ class Guide < ActiveRecord::Base
     editions.published.any?
   end
 
-  def can_be_unpublished?
+  def has_been_published?
     has_any_published_editions? && !has_any_unpublished_editions?
   end
 
--- a/app/views/editions/_actions.html.erb
+++ b/app/views/editions/_actions.html.erb
@@ -4,8 +4,12 @@
 
       <% edition_policy = EditionPolicy.new(current_user, edition) %>
 
-      <% if edition_policy.can_preview? %>
+      <% if edition_policy.can_preview? && edition_policy.can_discard_draft? %>
         <%= link_to "Preview", preview_content_model_url(guide), class: 'btn btn-default add-right-margin' %>
+      <% end %>
+
+      <% if guide.can_be_unpublished? %>
+        <%= link_to "View on website", [Plek.new.website_root, guide.slug].join(''), class: 'btn btn-default add-right-margin' %>
       <% end %>
 
       <% if !publish_controls_only %>
--- a/app/views/editions/_actions.html.erb
+++ b/app/views/editions/_actions.html.erb
@@ -4,7 +4,7 @@
 
       <% edition_policy = EditionPolicy.new(current_user, edition) %>
 
-      <% if edition_policy.can_preview? && edition_policy.can_discard_draft? %>
+      <% if edition_policy.can_preview? %>
         <%= link_to "Preview", preview_content_model_url(guide), class: 'btn btn-default add-right-margin' %>
       <% end %>
 
--- a/app/views/editions/_actions.html.erb
+++ b/app/views/editions/_actions.html.erb
@@ -8,7 +8,7 @@
         <%= link_to "Preview", preview_content_model_url(guide), class: 'btn btn-default add-right-margin' %>
       <% end %>
 
-      <% if guide.can_be_unpublished? %>
+      <% if guide.has_been_published? %>
         <%= link_to "View on website", [Plek.new.website_root, guide.slug].join(''), class: 'btn btn-default add-right-margin' %>
       <% end %>
 
@@ -48,7 +48,7 @@
             %>
         <% end %>
 
-        <% if guide.can_be_unpublished? %>
+        <% if guide.has_been_published? %>
           <%= link_to "Unpublish", unpublish_guide_path(guide), class: 'btn btn-danger pull-right' %>
         <% end %>
 
--- a/app/policies/edition_policy.rb
+++ b/app/policies/edition_policy.rb
@@ -17,7 +17,7 @@ class EditionPolicy
   end
 
   def can_discard_draft?
-    !Edition::STATES_THAT_UPDATE_THE_FRONTEND.include?(edition.state)
+    is_being_edited?
   end
 
   def can_discard_new_draft?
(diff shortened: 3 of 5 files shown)


36degrees:
I think you have an errant `:w` in your commit message, probably from vim?

spec/features/guide_publishing_flow_spec.rb
```diff
@@ -255,10 +255,64 @@
       end
       expect(page).to_not have_button "Publish Guide"
       expect(page).to_not have_button "Send for review"
+      # expect(page).to have_link "View on website", href: [Plek.new.website_root, guide.slug].join('')
```
36degrees: Did you mean to add this?

spec/features/guide_publishing_flow_spec.rb
```diff
@@ -255,10 +255,64 @@
...
+
+        it "has a 'View on website' link" do
+          expect(page).to have_link "View on website", href: "http://www.dev.gov.uk/service-manual/topic-name/published-guide"
+        end
+      end
     end
   end
 
-private
+  private
```
36degrees: GDS house style is not to indent the private declaration.

spec/features/guide_publishing_flow_spec.rb
```diff
@@ -231,7 +231,7 @@
 
   describe "guide edition history" do
     it "allows seeing previous edition changes" do
-      guide = create(:guide, :with_published_edition, title: "Original Title")
+      guide = create(:guide, :with_published_edition, title: "Original Title", slug: "/service-manual/topic-name/original-title")
```
36degrees: I don't think this needs to change as you never do anything with this slug?

app/views/editions/_actions.html.erb
```diff
@@ -4,10 +4,14 @@
 
       <% edition_policy = EditionPolicy.new(current_user, edition) %>
 
-      <% if edition_policy.can_preview? %>
+      <% if edition_policy.can_preview? && edition_policy.can_discard_draft? %>
         <%= link_to "Preview", preview_content_model_url(guide), class: 'btn btn-default add-right-margin' %>
       <% end %>
 
+      <% if guide.can_be_unpublished? %>
```
36degrees: It might be good to rename this methods to e.g. `has_been_published?`

app/views/editions/_actions.html.erb
```diff
@@ -4,10 +4,14 @@
 
       <% edition_policy = EditionPolicy.new(current_user, edition) %>
 
-      <% if edition_policy.can_preview? %>
+      <% if edition_policy.can_preview? && edition_policy.can_discard_draft? %>
```
36degrees: Likewise, could this be something like `is_being_drafted`…?

spec/features/guide_publishing_flow_spec.rb
```diff
@@ -255,10 +255,64 @@
...
+        it "has a 'View on website' link" do
+          expect(page).to have_link "View on website", href: "http://www.dev.gov.uk/service-manual/topic-name/just-published"
+        end
+      end
+
+      context "that has one published guide edition" do
+        before do
+          guide = create(:guide, :with_previously_published_edition,
+            slug: "/service-manual/topic-name/published-guide",
+            )
```
36degrees: I tend to outdent closing parentheses

So:

```rb
guide = create(:guide, :with_previously_published_edition,
  slug: "/service-manual/topic-name/published-guide",
)
```

Rather than:


```rb
guide = create(:guide, :with_previously_published_edition,
  slug: "/service-manual/topic-name/published-guide",
  )
```

That might just be personal preference though. Don't know if @thehenster has an opinion on this?

spec/features/guide_publishing_flow_spec.rb
deborahchua: I think this happened when I reindented the file in Sublime... :(

app/views/editions/_actions.html.erb
deborahchua: I'm not sure about changing this as it breaks the method naming convention from all the others (they all start with 'can')...what do you think @thehenster ?

spec/features/guide_publishing_flow_spec.rb
thehenster: In theory - I don't really mind. If the linter doesn't care then it's not really important.

In reality - my preference probably changes depending on the size and shape of the particular bit of code.

app/views/editions/_actions.html.erb
thehenster: It would make sense to me if the logic went into the `#can_preview?` method. Keeping the confusion of conditions out of the view is a good thing.

I get where Ollie is going though. Whether a draft can be discarded or not doesn't seem directly related to whether it can be previewed. Perhaps a private method would make it clearer? Eg.

```
def can_discard_draft?
    is_being_edited?
  end

  def can_preview?
    edition.persisted? && is_being_edited?
  end

private

  def is_being_edited?
    !Edition::STATES_THAT_UPDATE_THE_FRONTEND.include?(edition.state)
  end
```

app/models/guide.rb
```diff
@@ -89,7 +89,7 @@ def has_any_published_editions?
     editions.published.any?
   end
 
-  def can_be_unpublished?
+  def has_been_published?
```
thehenster: Good rename :+1:

app/views/editions/_actions.html.erb
deborahchua: I like that! I will make this change.

thehenster:
I've done some work smashing spec/features/guide_spec.rb and spec/features/guide_publishing_flow_spec.rb together. I never understood the distinction.

If it's OK with you @deborahchua, I would prefer to merge my 'smash' before this gets merged to avoid the conflict with the added tests in spec/features/guide_publishing_flow_spec.rb? I'll do the rebase for this branch once it's done.

deborahchua:
All good with me!

thehenster:
Hi @deborahchua, I've moved the tests from the (no longer) publishing_flow_spec.rb. Look OK?

deborahchua:
Yup

PR #351, opened 2016-10-18
Destroy associations when destroying guides

deborahchua:


thehenster:
I think we noticed that `Approval`s aren't `dependent: :destroy`ed either, but thinking about it now, we won't ever destroy a guide that has an approval so all good.

Merge away!

36degrees:
I think we might allow a user to discard a draft that's been approved, which would do that? But I think that's probably worth doing in a separate PR…

PR #353, opened 2016-10-21
Rake task to output published guide edtion history

deborahchua:
Some of the guides do not have correct change notes as they were created before the history feature was added. This rake task outputs details of all the published guides and details about their major changes. This will allow us to locate which change notes need to be updated.


thehenster:
I've renamed the name of this task to be a bit more specific about the nature of the output we're printing for each guide.

36degrees:
Yep, good call. Team work!

PR #354, opened 2016-10-21
Only consider published guides when building list of related communities

36degrees:
When building the list of content owners, which is used to create the list of related communities for the topic page, we currently consider guides that may not have been published yet.

This can be confusing if the topic gets republished before the guide gets published, as the community will appear on the topic page before it actually gets associated through the publishing of the guide.

### DIFF
--- a/app/presenters/topic_presenter.rb
+++ b/app/presenters/topic_presenter.rb
@@ -55,7 +55,7 @@ private
   end
 
   def content_owner_content_ids
-    content_ids = topic.guides.map do |guide|
+    content_ids = topic.guides.live.map do |guide|
       edition = guide.latest_edition
 
       edition.content_owner.content_id if edition.content_owner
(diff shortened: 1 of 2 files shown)


thehenster:
Does this mean we won't be able to create a new guide without it's community page being published?

This might not be a huge issue but in the past the content team have got everything into a ready state and then published all at once.

Getting a publishable version of the content for the community page before writing a guide for that community might be tricky.

36degrees:
As far as I'm aware it shouldn't have any effect on when we can create guides. All it means is that when we save and publish a topic it'll only include the content owners for guides that have been published.

thehenster:
Sorry - I didn't read it carefully enough. I thought at a glance that query was affecting the list of available communities (or content owners) in the guide form dropdown. I think we'd just looked at it a moment earlier.

Looks good. :+1:

PR #356, opened 2016-10-31
Cannot update guide if changed

thehenster:
Most of this PR is a consolidation of the guide feature specs. I thought I was going to append to them for fixing this bug. In the end I didn't but it's still valuable stuff.

We encountered a bug where a guide was published while it was currently being edited by someone else. This resulted in an edition version getting published twice, which we want to avoid.

We have defended against this here but not allowing a guide to be altered if the guide has been changed since it was opened for editing in the browser. The unique bit of information that identifies the last edition, or fingerprint, is just the edition ID because editions are (in practise) immutable. We only create new editions rather than update them.

This should defend against other unexpected changes in state..

Trello ticket for the bug: https://trello.com/c/vzYnhuX4


app/controllers/guides_controller.rb
```diff
@@ -71,31 +71,28 @@ def confirm_unpublish
...
     message = opts.fetch(:message, nil)
 
+    @guide_form = GuideForm.build(
+      guide: guide,
+      edition: guide.latest_edition,
+      user: current_user
+    )
+
+    if guide_has_changed_since_editing?(guide)
+      flash.now[:error] = "The guide has changed since you started editing"
```
thehenster: The problem is that what they do is going to depend a bit on the type of change they've made.. They're in conflict resolution land. They might want to merge the changes with 'theirs' or 'mine' as a base.

We could talk about navigating back to the guides index, reopening the guide they're editing, comparing the differences, and leaving a question mark over how to proceed. Sounds like a bit of a long flash message :/ Could be done though.

36degrees: Maybe something to show and tell with the content designers, see if they understand it?

thehenster: Aye. I'll remove the reference to 'guide'.

thehenster: I had a go at improving it..

PR #357, opened 2016-11-11
Allow a guides topic to change until it is published

36degrees:
https://trello.com/c/CgDRSwG8/440-allow-a-guide-s-topic-to-be-changed-until-it-is-published

### DIFF
--- a/spec/models/guide_spec.rb
+++ b/spec/models/guide_spec.rb
@@ -61,84 +61,76 @@ RSpec.describe Guide do
   end
 
   describe "validations" do
-    it "ensures that slugs aren't saved without the topic name in the path" do
-      guide = Guide.new(slug: "/service-manual/guide-path")
-      guide.valid?
-      expect(guide.errors.full_messages_for(:slug)).to eq ["Slug must be present and start with '/service-manual/[topic]'"]
+    describe "the slug" do
+      it "is not valid when it does not include a topic" do
+        guide = build(:guide, slug: "/service-manual/guide-path")
+
+        expect(guide).not_to be_valid
+        expect(guide.errors.full_messages_for(:slug)).to eq [
+          "Slug must be present and start with '/service-manual/[topic]'"
+        ]
+      end
+
+      it "is not valid when it includes non alphanumeric characters" do
+        guide = build(:guide, slug: "/service-manual/topic-name/$")
+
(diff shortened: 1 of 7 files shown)


spec/models/guide_spec.rb
```diff
@@ -60,115 +60,129 @@
...
           topic: create(:topic, path: "/service-manual/different-topic")
         )
 
-        guide = create(:guide,
+        guide = create(:guide, :with_published_edition,
           topic_section:  original_topic_section
         )
+
         guide.topic_section_guides[0].topic_section_id = different_topic_section.id
+        expect(guide).not_to be_valid
```
thehenster: Calling `#valid?` and then `#save` will rerun the validations. Maybe `expect(guide.save).to eq(false)`? It might be possible to drop the `guide.save` instead, but as this test is explicitly reloading stuff from the database I'd probably leave the save in there..

spec/models/guide_spec.rb
```diff
@@ -60,115 +60,129 @@
...
-        expect(original_topic_section.reload.guides).to_not include guide
+      it "is possible to change the slug of a never published guide" do
+        guide = create(:guide, :with_draft_edition)
+
+        guide.slug = "/service-manual/topic-name/something-else"
+
+        expect(guide).to be_valid
+      end
+
+      it "is not possible to change the slug of a published guide" do
```
thehenster: I much prefer this description to the context block below :+1:

spec/models/guide_spec.rb
```diff
@@ -60,115 +60,129 @@
     end
   end
 
-  describe "validations" do
-    it "ensures that slugs aren't saved without the topic name in the path" do
-      guide = Guide.new(slug: "/service-manual/guide-path")
-      guide.valid?
-      expect(guide.errors.full_messages_for(:slug)).to eq ["Slug must be present and start with '/service-manual/[topic]'"]
-    end
+  describe "#valid?" do
```
thehenster: The `describe "validations"` isn't essential but it's pretty common. The validations are also run when you call `#save`/ `#create`/`#update`. Neither is really wrong but I think most of this app does `describe "validations"` and talks about the thing it's validating in another nested block. eg.

```
describe "validations" do
  describe "slug" do
    it "does this"
    it "does that"
  end
end
```

PR #358, opened 2016-11-22
Allow redirecting to topic sections (by anchor)

36degrees:
Break topics down to include ‘anchor’ link options for all sub topics as well as a link option for the topic itself.

screen shot 2016-11-22 at 10 16 50

This allows us to redirect old parts of the manual to a specific sub topic within a topic, even though it has no page of its own.

### DIFF
--- a/app/helpers/redirect_destination_helper.rb
+++ b/app/helpers/redirect_destination_helper.rb
@@ -14,7 +14,7 @@ private
   end
 
   def topic_select_options
-    Topic.order(:path)
+    Topic.includes(:topic_sections).order(:path)
       .flat_map { |topic| paths_for_topic_and_its_sections(topic) }
   end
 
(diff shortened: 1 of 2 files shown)


app/helpers/redirect_destination_helper.rb
```diff
@@ -1,16 +1,28 @@
...
   end
+
+private
+
+  def guide_select_options
+    Guide.live.order(:slug).pluck(:slug)
+  end
+
+  def topic_select_options
+    Topic.order(:path)
```
mcmillan: It might be worth adding an `includes(:topic_sections)` to this; without it, we'll be making an extra query for each topic in `paths_for_topic_and_its_sections`.

36degrees:
@mcmillan Thanks! I've made a couple of changes based on your comments – can you please take a second look?

mcmillan:
@36degrees LGTM :shipit:

PR #363, opened 2016-12-20
Add more redirects from the old service manual

36degrees:


36degrees:
For reference, these were originally redirected from within `design-principles` : https://github.com/alphagov/design-principles/blob/61c8446c14ac706e2f8724ed4736353a515797df/config.ru

PR #368, opened 2017-01-09
Add change note migrator

mcmillan:
This adds a `ChangeNoteMigrator` class, which is designed to enable us
to update the change notes on existing editions where en masse where
they've been populated with the wrong data erroneously.

In short, it updates change notes for an edition, or makes an edition
minor or major, then republishes via the `GuideRepublisher` class,
providing some pretty console output along the way.

Also included is an example rake task which can be used for running the
thing.

### DIFF
--- a/spec/models/change_note_migrator_spec.rb
+++ b/spec/models/change_note_migrator_spec.rb
@@ -7,80 +7,98 @@ RSpec.describe ChangeNoteMigrator do
   let(:minor_edition) { create(:edition, :published, update_type: "minor", version: 2, guide: guide) }
   let(:unpublished_edition) { create(:edition, :draft, guide: guide) }
   let(:change_note) { "I am a change note" }
-  subject { described_class.new(publishing_api: publishing_api, dry_run: false) }
 
-  before do
-    allow(publishing_api).to receive(:put_content)
-    allow(publishing_api).to receive(:patch_links)
-    allow(publishing_api).to receive(:publish)
-  end
+  context "with dry_run mode enabled" do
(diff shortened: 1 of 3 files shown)


spec/models/change_note_migrator_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+
+    it "updates the edition's change note" do
+      subject.update_change_note(major_edition.id, change_note)
+      major_edition.reload
+      expect(major_edition.change_note).to eq(change_note)
+    end
+
+    it "updates the guide in the publishing API" do
+      aggregate_failures do
```
36degrees: TIL. Worth naming these aggregations?

spec/models/change_note_migrator_spec.rb
```diff
@@ -0,0 +1,86 @@
+require "rails_helper"
+
+RSpec.describe ChangeNoteMigrator do
+  let(:publishing_api) { double(:publishing_api) }
+  let(:guide) { create(:guide) }
+  let(:major_edition) { create(:edition, :published, update_type: "major", version: 2, guide: guide) }
+  let(:minor_edition) { create(:edition, :published, update_type: "minor", version: 2, guide: guide) }
+  let(:unpublished_edition) { create(:edition, :draft, guide: guide) }
+  let(:change_note) { "I am a change note" }
+  subject { described_class.new(publishing_api: publishing_api, dry_run: false) }
```
36degrees: Part of me thinks this could use some tests for dry run behaviour. Part of me thinks this might only be used once… up to you.

PR #371, opened 2017-01-10
Tiny edit on style guide desc

keithiopia:
Changed 'gov.uk' to 'government' as it helps our colleagues out in depts to argue the use of it


36degrees:
Thanks! 👍

PR #377, opened 2017-01-18
Use Jenkins Pipeline

36degrees:
This commit adds a Jenkinsfile which allows us to move onto the new Jenkins 2 CI infrastructure. It is based on the Jenkinsfile for government-frontend, but with slightly simplified stages.

This will replace the existing jenkins.sh file once we’re happy it works.


Jenkinsfile
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env groovy
+
+REPOSITORY = 'service-manual-publisher'
+
+// @todo Replace with govuk.setupDb when on Rails 5.
+def setupDb() {
```
mcmillan: On the off-chance that we remain on Rails 4.x for longer than anticipated, can we add a comment to explain _why_ we need this because we're not on Rails 5?

36degrees:
What is it you think is missing from the setupDb documentation?

mcmillan:
@36degrees Sorry – my line note appears to have gone walkabout.

It's a very pedantic comment, so feel free to ignore, however my thinking was that it might be worth explaining why we're using our own version of `setupDb` in case it ends up lingering around for longer than anticipated; I'm not sure what our timeline is for migrating to Rails 5.

PR #380, opened 2017-01-31
Remove reason for change

mcmillan:
This PR:
- Removes the "reason for change" field from editions
- Updates the change note UI when editing a guide to match Whitehall Publisher


36degrees:
Think there's a [rake task](https://github.com/alphagov/service-manual-publisher/blob/master/lib/tasks/print_change_history_for_published_guides.rake#L13) you might wanna fix too.

PR #381, opened 2017-01-31
Update README to describe creating a user

chrisroos:
We need at least one user in the database to run this application
locally.

I've copied the text from the manuals-publisher README[1].

[1]:
https://github.com/alphagov/manuals-publisher#running-the-application

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,13 @@ bundle exec rake db:setup
 
 ### Development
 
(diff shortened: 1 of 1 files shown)


chrisroos:
Thanks for the review, @cbaines. I rebased the branch on master, force pushed and have now merged.

PR #388, opened 2017-02-13
[DO NOT MERGE] Remove slug migrations

36degrees:
Waiting for confirmation from content designers that they're done with the migrations

https://trello.com/c/tiHcWYpt/579-remove-slug-migration-functionality-from-publisher


db/migrate/20170213132040_remove_slug_migrations.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveSlugMigrations < ActiveRecord::Migration[5.0]
+  def up
+    drop_table :slug_migrations
+  end
+
+  def down
+    fail ActiveRecord::IrreversibleMigration
```
mcmillan: Does this have to be irreversible?

36degrees: I was debating making it reversible, but once the data's gone it seemed like a bit of a moot point…?

36degrees:
Closing this as we've had feedback from the content designers that as the manual evolves they may want to alter previous migrations to go somewhere that's a better fit. So instead of this, we'll need to alter the migrations behaviour to allow them to be changed – as currently migrations that are complete are not modifiable.

PR #389, opened 2017-02-13
Topic form improvements

36degrees:
- Re-order fields to match the guide form
- Consistently use 'Slug' and generate 'Final URL' rather than requiring an entire path
- Automatically generate the slug from the title
- Use auto expanding text area for the description rather than text field

## Before
screen shot 2017-02-13 at 15 40 35

## After
screen shot 2017-02-13 at 15 40 48

https://trello.com/c/ED6wNNn5/444-re-order-top-fields-of-the-topics-edit-pages-in-the-pub-app


36degrees:
Thanks for the review @cbaines!

PR #392, opened 2017-02-17
[Do Not Merge] Add poster_url to ServiceStandardPresenter

mcmillan:
This needs to be changed to the real URL before merging.


36degrees:
Tempted just to close this for now, until we have the poster - given pretty much 100% of the changes will be squashed into something new anyway.

mcmillan:
@36degrees Yeah, let's do that.

PR #396, opened 2017-06-21
Fix bug where clicking in empty section removes it

36degrees:
Somehow this invalid HTML (closing a link as if it’s a button) causes a weird issue where clicking into the title field of the section causes it to be removed from the DOM.


gemmaleigh:
👍

PR #398, opened 2017-07-26
Send update_type to put_content

emmabeynon:
We are deprecating sending the `update_type` in publish requests made to the Publishing API, and now expect clients to send `update_type` in PUT requests instead. We also upgrade gds-api-adapters to bring this app up to date with the new expected behaviour.

https://trello.com/c/LnJlkb9e/987-5-deprecate-the-usage-of-updatetype-in-publish-for-publishing-api-and-update-usage-across-govuk

### DIFF
--- a/app/models/guide_republisher.rb
+++ b/app/models/guide_republisher.rb
@@ -12,7 +12,7 @@ class GuideRepublisher
 
       publishing_api.put_content(guide.content_id, guide_for_publication.content_payload)
       publishing_api.patch_links(guide.content_id, guide_for_publication.links_payload)
-      publishing_api.publish(guide.content_id, "minor")
+      publishing_api.publish(guide.content_id)
     end
   end
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -82,7 +82,7 @@ GEM
     diff-lcs (1.3)
     diffy (3.1.0)
     docile (1.1.5)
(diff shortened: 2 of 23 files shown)


app/models/guide_republisher.rb
```diff
@@ -12,7 +12,7 @@ def republish
 
       publishing_api.put_content(guide.content_id, guide_for_publication.content_payload)
```
36degrees: Should this have had 'minor' moved to here, or am I not understanding this change?

thomasleese: It's available in `guide_for_publication.content_payload`.

36degrees: But this script is meant to re-publish guides but always as a minor edition, as we use this script if e.g. we need to change the way a guide is presented, but we don't want to trigger email alerts or show a change in the history of the guide.

thomasleese: Ah sure, we'll revert this change and use the `republish` `update_type` here.

36degrees: Thanks, appreciate it 👍

emmabeynon: Thanks for explaining @36degrees - PR is open to fix this here https://github.com/alphagov/service-manual-publisher/pull/399

PR #401, opened 2017-07-27
Update responses to behave like a hash

emmabeynon:
Following on from the gds-api-adapters upgrade to 47.2, OpenStruct responses
are now longer supported and must now be hashes.


36degrees:
Thank you!

PR #404, opened 2017-09-19
Migrate from Airbrake to Sentry

36degrees:
Following the instructions on https://docs.publishing.service.gov.uk/manual/upgrade-to-sentry.html

https://trello.com/c/6LG9tlZN/66-update-service-manual-publisher


36degrees:
@danielroseman would you mind reviewing those extra two commits? Ta!

PR #408, opened 2018-02-09
Handle external urls

cbaines:


cbaines:
Thanks for taking a look Simon :)

PR #486, opened 2018-06-06
Add primary publishing organisation to each new or updated guide

KyleMacPherson:
Add a primary publishing org field to links so that we can track primary orgs in the publishing api


KyleMacPherson:
[(2) Set primary org in Service Manual Publisher](https://trello.com/c/LnvPOxjY/379-2-set-primary-org-in-service-manual-publisher)

PR #499, opened 2018-07-02
Service Toolkit - update

TechPunk316:
Additions to standards and guidance based on user research
Update to Design System link


36degrees:
Syntax errors - fixed in #500

PR #530, opened 2018-09-03
Rename standard

benjamin-mortimer:
The Digital Service Standard is being renamed the Government Service Standard and being reduced to 14 points.

The title and description of the Standard aren't things the Service Manual content team can change using the Service Manual Publisher interface.

This PR changes the title and description where they're set in Service Manual Publisher.

It'll probably need more work.

### DIFF
--- a/app/presenters/homepage_presenter.rb
+++ b/app/presenters/homepage_presenter.rb
@@ -9,7 +9,7 @@ class HomepagePresenter
     {
       base_path: '/service-manual',
       title: 'Service Manual',
-      description: 'Helping government teams create and run great digital services that meet the Government Service Standard.',
+      description: 'Helping government teams create and run great digital services that meet the Service Standard.',
       details: {},
       routes: [
         { type: 'exact', path: '/service-manual' }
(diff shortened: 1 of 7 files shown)


benjamin-mortimer:
@36degrees - I've had a go at this but it needs a developer to look at.

The background is we need to change the name of the Service Standard, which is referenced throughout the Service Manual in various places like breadcrumbs and on 'point' pages.

We initially thought we could make this change in Service Manual Frontend by tweaking the templates, for example [the 'guide' template](https://github.com/alphagov/service-manual-frontend/blob/master/app/views/content_items/guide.html.erb) that's used to display points in the standard to users.

However, it's more complicated than that. For example, each point has the text 'Digital Service Standard' above its h1. We need to change this to 'Government Service Standard' to avoid confusion. That text isn't hard-coded in the template but pulled in from data in the content item: https://github.com/alphagov/service-manual-frontend/blob/master/app/views/content_items/guide.html.erb#L19.

As far as I know that data comes from the Content Store, and it gets to the Content Store from this presenter in Service Manual Publisher - hence making the change here. 

What do you think? Is this anything like the right approach?

36degrees:
Going to close this and re-open from a branch on the repo, as CI doesn't build forks.

PR #633, opened 2019-04-30
[DO NOT MERGE] Rename the service standard

36degrees:
This is an updated version of #530 – CI doesn't build forked branches, so had to open a new PR.

From the original PR:

> The Digital Service Standard is being renamed the Government Service Standard [UPDATE: now 'Service Standard'] and being reduced to 14 points.
>
> The title and description of the Standard aren't things the Service Manual content team can change using the Service Manual Publisher interface.
>
> This PR changes the title and description where they're set in Service Manual Publisher.
>
> It'll probably need more work.


36degrees:
Squashed down to one commit.

PR #814, opened 2020-06-03
Actually support republishing

benthorner:
https://trello.com/c/KNWv9Dgx/258-migrate-legacy-email-signup-pages-for-service-manual-formats

Previously we had a "republish" rake task that would publish
a "major" change when invoked. Since a "major" change becomes
part of the content history, and can lead to email alerts, we
should try to use a "republish" update type instead. This does
that, consolidates the existing code, and makes it more flexible.

Please see the commits for more details.


benthorner:
Update: the inaccurate `email_alert_signup` links, for topics and the service standard, seem to have vanished by themselves. This is probably because the linked content items are unpublished, and any other publishing that triggered a link expansion will automatically compacted them away.

However, they are still present in the Publishing API, so we should still republish the content items to remove the redundant links. I've added a commit to support doing that.

app/presenters/service_standard_presenter.rb
```diff
@@ -30,6 +30,7 @@ def links_payload
     {
       links: {
         primary_publishing_organisation: [ServiceManualPublisher::GDS_ORGANISATION_CONTENT_ID],
+        email_alert_signup: [],
```
chao-xian, line 33: Is it worth having this in a separate PR to revert? Doesn't feel like it supports republishing really.

benthorner: Sure, I've split that out: https://github.com/alphagov/service-manual-publisher/pull/815

chao-xian: 👍

PR #845, opened 2020-08-27
Bump govspeak and govuk_publishing_components

kevindew:
Trello: https://trello.com/c/H49ZOsu1/972-update-all-govspeak-depenendencies-in-govuk-apps

The govspeak bump resolves
[CVE-2020-14001](https://github.com/advisories/GHSA-mqm2-cgpr-p4m6).

The govuk_publishing_components bump is quite significant and introduces
some govuk-frontend SASS that causes problems for the sprockets approach
to compressing SASS (it uses the SASS processor to first convert the
SASS to CSS, and then runs it through it again to compress that CSS).
This does not work for SASS functions that conflict with CSS native
functions such as: `unquote("max(calc(0px))"` [1].

To resolve this we need to disable sprockets CSS compressor and instead
use SASS itself to do the compression.

[1]: https://github.com/alphagov/govuk-frontend/blob/9dc2cfb064fb49e421d5a73be6a241506e1414b8/src/govuk/objects/_width-container.scss#L46


gclssvglx (approved):
LGTM 👍

PR #847, opened 2020-09-07
Update bundle with ruby 2.7.1

benbaumann95:
- These are the necessary changes for Ruby 2.7
- Since there is an automated process for updating Ruby versions in govuk apps, the ruby version has been changed locally to 2.7.1 with the necessary updates, but a PR raised with version 2.6.6. The tests will fail on Jenkins because of this.
- No deprecation warnings, just the gems and bundler update.


cdccollins (approved):
Just need to work out what to do about the Bundler version

PR #875, opened 2020-12-08
Add Healthcheck for ActiveRecord

MuriloDalRi:
This healthcheck is a prerequisite for enabling continous deployment
for this app.

### DIFF
--- a/spec/requests/healthcheck_spec.rb
+++ b/spec/requests/healthcheck_spec.rb
@@ -1,0 +1,8 @@
+RSpec.describe "/healthcheck" do
+  it "returns database connection status" do
+    get "/healthcheck"
+    json = JSON.parse(response.body)
+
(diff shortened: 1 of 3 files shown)


spec/requests/healthcheck_spec.rb
```diff
@@ -0,0 +1,8 @@
+RSpec.describe "/healthcheck" do
+  it "returns database connection status" do
+    get "/healthcheck"
+    json = JSON.parse(response.body)
+
+    expect(json["checks"]).to include("database_connectivity")
```
ChrisBAshton, line 6: Probably worth testing the status is OK and/or the HTTP response code is 200, like the other apps have been doing. Otherwise this is looking great 👍

MuriloDalRi: Done

ChrisBAshton (approved):
Thanks

PR #960, opened 2021-07-26
Fix Rails data migration (and Docker container initialisation)

ChrisBAshton:
Currently, attempting to make service-manual-publisher in govuk-docker is broken as the Rails data migration cannot complete. This PR fixes that - see commits for details.

Related Trello: https://trello.com/c/nWRFaLO3/2562-manually-bump-apps-to-rails-61

### DIFF
--- a/db/migrate/20160510122323_index_foreign_keys_in_approvals.rb
+++ b/db/migrate/20160510122323_index_foreign_keys_in_approvals.rb
@@ -1,6 +1,6 @@
 class IndexForeignKeysInApprovals < ActiveRecord::Migration[5.2]
   def change
-    add_index :approvals, :edition_id
-    add_index :approvals, :user_id
+    add_index :approvals, :edition_id unless index_exists?(:approvals, :edition_id)
+    add_index :approvals, :user_id unless index_exists?(:editions, :user_id)
   end
 end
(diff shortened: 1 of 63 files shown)


db/migrate/20160510122323_index_foreign_keys_in_approvals.rb
```diff
@@ -1,6 +1,6 @@
-class IndexForeignKeysInApprovals < ActiveRecord::Migration
+class IndexForeignKeysInApprovals < ActiveRecord::Migration[5.2]
   def change
-    add_index :approvals, :edition_id
-    add_index :approvals, :user_id
+    add_index :approvals, :edition_id unless index_exists?(:approvals, :edition_id)
+    add_index :approvals, :user_id unless index_exists?(:editions, :user_id)
```
benjamineskola, line 4: Should this be `:approvals` not `:edition`?

When trying to apply it I get an error because `index_approvals_on_user_id` already exists.

ChrisBAshton: Good spot - it should be! 🤦

ChrisBAshton:
```suggestion
    add_index :approvals, :user_id unless index_exists?(:approvals, :user_id)
```

benjamineskola:
I've rebased #952 on top of this but still encounter the same errors — I'm not entirely sure yet whether migrations are being run correctly or if there's some missing step still.

benjamineskola:
I wonder if [this change](https://github.com/alphagov/service-manual-publisher/pull/952/commits/4de817ba756ba482d70abba22b2fb68e0e75e52b) actually would be better placed in this PR — it's not really part of the 6.1 upgrade but coincidental. Or perhaps a third PR in between these two?

PR #962, opened 2021-08-03
GOVSI-484: Add "GOV.UK Sign in" to the service toolkit

chrisclayson:
Digital Identity are starting a beta programme for the GOV.UK Sign in product. Add the link to the product page to the service toolkit.

### DIFF
--- a/app/presenters/service_toolkit_presenter.rb
+++ b/app/presenters/service_toolkit_presenter.rb
@@ -143,11 +143,6 @@ class ServiceToolkitPresenter
             "description": "Take and process payments - a simple experience for users and easy integration for you",
           },
           {
-            "title": "GOV.UK Verify",
-            "url": "https://www.verify.service.gov.uk",
-            "description": "Let users prove their identity to you securely and conveniently",
-          },
-          {
             "title": "GOV.UK Platform as a Service",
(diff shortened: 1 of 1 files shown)


sanisc (approved):

StephenGill:
Hi both - I wonder if there's a way to make it clearer that the service isn't live yet, and the call to action here is to to register for the private beta. Something like -

```

{
            "title": "GOV.UK Sign in (beta)",
            "url": "https://www.sign-in.service.gov.uk",
            "description": "Let users sign in to your service quickly, easily and securely - register for the private beta",
          },

```

Let me know what you think.

chrisclayson:
> Hi both - I wonder if there's a way to make it clearer that the service isn't live yet, and the call to action here is to to register for the private beta. Something like -
> 
> ```
> 
> {
>             "title": "GOV.UK Sign in (beta)",
>             "url": "https://www.sign-in.service.gov.uk",
>             "description": "Let users sign in to your service quickly, easily and securely - register for the private beta",
>           },
> ```
> 
> Let me know what you think.

Good point @StephenGill ... we have updated appropriately.

sanisc (approved):

CharlesIC (approved):

alex-ju (approved):

PR #1008, opened 2022-02-03
Use schema.rb supplemented by custom_functions.sql

ollietreend:
This commit attempts to reconcile work done in 4eff0b9afbcc70ae75b376329725c5e45255e19d and bc6d9da55680acb22b845c1703af265e67c27e88.

The aim is to use the usual `schema.rb` file for _most of_ the app's database schema. This is the conventional schema format for Rails apps, and is consistent with existing GOV.UK apps.

Additionally, there's a custom function and trigger which need to be defined in the database. The `schema.rb` format cannot define custom functions – the only way to achieve this is with a custom SQL file.

Now when executing `rake db:setup` (or similar), the primary `schema.rb` file will be imported. Then the file `db/custom_functions.sql` will automatically be imported to supplement the database schema.


benjamineskola (approved):
Looks good to me!

PR #1019, opened 2022-03-01
Migrate jasmine tests to jasmine-browser-runner

callumknights:
Jasmine Ruby gems are deprecated and no longer compatible with Rails 7, which is blocking upgrades.

Trello: https://trello.com/c/21VMx1NV/264-migrate-apps-to-use-jasmine-browser-runner


Gemfile
```diff
@@ -31,8 +31,6 @@ group :development, :test do
   gem "byebug"
   gem "fuubar"
   gem "govuk_test"
-  gem "jasmine"
```
kevindew, line 34: We should remove jasmine from the dependabot configuration: https://github.com/alphagov/service-manual-publisher/blob/849d2a77ecae7ea2357085c10b218a5d00e91828/.github/dependabot.yml#L24-L25

package.json
```diff
@@ -28,6 +31,8 @@
     "extends": "stylelint-config-gds/scss"
   },
   "devDependencies": {
+    "jasmine-browser-runner": "^1.0.0",
+    "jasmine-core": "^4.0.1",
```
kevindew, line 35: We should add these two to the NPM dependabot configuration: https://github.com/alphagov/service-manual-publisher/blob/849d2a77ecae7ea2357085c10b218a5d00e91828/.github/dependabot.yml#L43-L47

kevindew (commented):
This looks good, I think we just need the dependabot configuration to finish it.

It'd be good if this could have a commit message to explain why we've done this.

Gemfile
callumknights: 👍

package.json
callumknights: 👍

benjamineskola (approved):

PR #1026, opened 2022-03-09
Upgrade to Rails 7

callumknights:
Upgrading to Rails 7

Trello: https://trello.com/c/JAycjHYZ/252-upgrade-service-manual-publisher-from-rails-61x-to-70x


Gemfile
```diff
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem "rails", "6.1.5"
+gem "rails", "7.0.2"
```
brucebolt, line 3: [Version 7.0.2.3](https://rubygems.org/gems/rails/versions/7.0.2.3) is now available, would it be possible to bump to that version now?

config/environments/production.rb
```diff
@@ -100,30 +82,4 @@
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
-
-  config.action_mailer.delivery_method = :notify
-  config.action_mailer.notify_settings = {
-    api_key: Rails.application.secrets.notify_api_key,
-  }
```
brucebolt, line 107: I think you'll need to keep these lines, as they tell the application to use GOV.UK Notify to send emails, instead of the default mailer.

Gemfile
```diff
@@ -19,6 +19,7 @@ gem "redcarpet"
 gem "rinku", require: "rails_rinku"
 gem "sassc-rails"
 gem "select2-rails"
+gem "sprockets-rails"
```
brucebolt, line 22: Could sprockets be added in a separate commit, so we have a commit message explaining why it was added as a dependency?

config/application.rb
```diff
@@ -23,7 +23,7 @@ module ServiceManualPublisher
 
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 6.1
+    config.load_defaults 7.0
```
brucebolt, line 26: Once this version number has been bumped here, the entire `config/initializers/new_framework_defaults_7_0.rb` file can then be removed, as Rails will automatically load the configuration that you uncommented in the defaults file.

brucebolt (commented):
Great work @callumknights.  I've added a few small comments.

Gemfile
callumknights: Annoyingly I actually did have this in a separate commit and then squashed it cause I wasn't sure it seemed worth leaving separate, I can rewrite that if necessary though.

brucebolt (approved):
Thanks for making those changes.  Everything looks good to me 👍 .

PR #1070, opened 2022-06-09
Add GOV.UK Forms information to Toolkit presenter

DavidBiddle:
In keeping with the other Digital Service Platforms products, we'd like the newly deployed [GOV.UK Forms product page](https://www.forms.service.gov.uk) to be linked to from the Service Toolkit. 

My understanding is that this PR will accomplish that. Please let me know if that's incorrect - I haven't been able to test this locally.

### DIFF
--- a/app/presenters/service_toolkit_presenter.rb
+++ b/app/presenters/service_toolkit_presenter.rb
@@ -152,6 +152,11 @@ class ServiceToolkitPresenter
             "url": "https://www.sign-in.service.gov.uk",
             "description": "Let users sign in to your service quickly, easily and securely. Register for the private beta.",
           },
+          {
+            "title": "GOV.UK Forms (private beta)",
+            "url": "https://www.forms.service.gov.uk",
+            "description": "Create accessible online forms for GOV.UK with no need for technical or design skills.",
+          },
         ],
       },
       {


36degrees:
As far as I know, once this has been merged and deployed someone with the right access will likely need to run the '[republish service toolkit](https://github.com/alphagov/service-manual-publisher/blob/main/lib/tasks/republish.rake#L34-L38)' rake task to get it to actually appear.

kevindew:
Yup, this looks good. 

I've given it a try and it'll output like this:
Screenshot 2022-06-15 at 09 49 31

My only query before merging is how signed off the copy is for this? I'm not sure if we have anyone who owns this service toolkit page any more so I guess the check would be that this is copy is happily signed off by senior stakeholders in Forms?

DavidBiddle:
Thanks Kevin - this text has been approved by our senior content designer and is equivalent to existing text on our product page which has previously been approved by our product manager, so we're happy for this to be deployed.

kevindew (approved):

DavidBiddle:
Thank you!

kevindew:
It's live now on https://www.gov.uk/service-toolkit (slight chance you may see a cached version for next 5 mins)

PR #1114, opened 2022-09-14
Dockerised app to meet replatforming standards.

nsabri1:
Adding Dockerfile which is preet standard we use for Ruby Apps Added .dockerignore to keeps image clean

https://trello.com/c/UW3kUhcG/728-package-service-manual-publisher-for-migration

removal of gds_api.rb to honour the RAILS_LOG_TO_STDOUT environment variable for GdsApi client logs.
https://github.com/alphagov/publisher/pull/1603


Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+ARG builder_image=ghcr.io/alphagov/govuk-ruby-builder:2.7.6
+ 
+FROM $builder_image AS builder
+
+WORKDIR /app
+
+COPY Gemfile* .ruby-version /app/
+
+RUN apt update && apt install libpq-dev -y && \
+    bundle install
```
sengi, line 11: Couple of minor improvements:
* use `--no-install-recommends` so as not to install unnecessary packages (this is mainly to reduce noise from vulnerability monitoring tools, but also helps with image size of course)
* run the `bundle install` separately from the package installation, to take better advantage of layer cachine.
```suggestion
RUN apt update && apt install -y --no-install-recommends libpq-dev
RUN bundle install
```

Normally we'd also do `&& rm -rf /var/lib/apt/lists/*` as part of the `RUN apt-get` instruction, but since this is the builder image it's fine not to. (https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get)

Dockerfile
```diff
@@ -0,0 +1,29 @@
+ARG base_image=ghcr.io/alphagov/govuk-ruby-base:2.7.6
+ARG builder_image=ghcr.io/alphagov/govuk-ruby-builder:2.7.6
+ 
+FROM $builder_image AS builder
+
+WORKDIR /app
+
+COPY Gemfile* .ruby-version /app/
+
+RUN apt update && apt install libpq-dev -y && \
```
sengi, line 10: We should probably just include libpq-dev in the builder image, but this is fine for now. Maybe add a TODO?

sengi (approved):

Dockerfile
sengi: Actually I just added libpq-dev to the builder image — much cleaner that way. https://github.com/alphagov/govuk-ruby-images/pull/11

So this can just become `RUN bundle install` here.

sengi (approved):

Dockerfile
sengi: I added it

sengi (approved):
LGTM. There's [prior](https://github.com/alphagov/frontend/commit/02fe98ab645f23dfdcf7b01c16f7402635cba4db) [art](https://github.com/alphagov/smart-answers/commit/eafb5785ffa9db9d3422087e0260486260474499) for removing the gdsapi logger so that's all good.

